PR #2, opened 2014-11-02
Add .pylintrc and runtests.

stephen-soltesz:
This pull request adds two files. The first is a pylintrc, which helps enforce style rules with pylint. The second is runtests, which is a simple bash script that automates running pylint, unit tests, and test-coverage reports based on the unit tests.

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -1,0 +1,241 @@
+[MASTER]
+
+# Specify a configuration file.
+#rcfile=
+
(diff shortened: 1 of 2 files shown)


stephen-soltesz:
A pre commit hook for linting .py files (at least) sounds good to me. The hooks are really flexible, and I'd like to experiment a little. I'll commit the current files in this pull request since it sounded like you were ok with them and send new hooks in the future. Please let me know if you had a different idea in mind.

PR #8, opened 2015-01-25
Add mlab_export_test.py and testdata.

stephen-soltesz:
This change includes tests for the mlab_export.py command line tool.

### DIFF
--- a/export/mlab_export_test.py
+++ b/export/mlab_export_test.py
@@ -177,8 +177,8 @@ class MlabExport_GlobalTests(unittest.TestCase):
   def testunit_assert_start_and_end_times(self):
     mock_options = mock.Mock()
     mock_options.length = 3600
-    mock_options.ts_end = 1401401409 + mock_options.length
-    mock_options.ts_start = 1401401409
+    mock_options.ts_end = FAKE_TIMESTAMP + mock_options.length
+    mock_options.ts_start = FAKE_TIMESTAMP
 
     try:
       returned_value = mlab_export.assert_start_and_end_times(mock_options)
--- a/export/mlab_export_test.py
+++ b/export/mlab_export_test.py
@@ -340,51 +340,56 @@ class MlabExport_GlobalTests(unittest.TestCase):
 
     mock_get_canonical_names.assert_called_with('file1', 'value', mock_options)
 
-  # TODO: account for make_output_dirs @mock.patch('mlab_export.os.makedirs')
+  @mock.patch('mlab_export.make_output_dirs')
   @mock.patch('mlab_export.get_rrd_files')
   @mock.patch('mlab_export.rrdtool.fetch')
   @mock.patch('mlab_export.get_canonical_names')
   @mock.patch('gzip.open')
   @mock.patch('StringIO.StringIO.close')
   def testunit_rrd_export(
-      self, mock_close, mock_open, mock_get_canonical_names, mock_rrdtool_fetch,
-      mock_get_rrd_files):
+      self, mock_close, mock_open, mock_get_canonical_names,
+      mock_rrdtool_fetch, mock_get_rrd_files, mock_make_output_dirs):
     str_fd = StringIO.StringIO()
     mock_open.return_value = str_fd
     mock_get_rrd_files.return_value = ['file1']
     mock_rrdtool_fetch.return_value = [
-        (0, 10, 10), ['value', 'skipped_metric'], [(0.0,0.0)]]
+        (0, 10, 10), ['value', 'unknown_metric', 'ignored_metric'],
+        [(0.0, 1.0, 2.0)]]
     mock_get_canonical_names.side_effect = [
-        ('fake_hostname', 'fake_experiment', 'fake_metric'),
-        ('fake_hostname', 'fake_experiment', None)]
+        ('fake_hostname', 'fake_experiment', 'fake_metric_name'),
+        ('fake_hostname', 'fake_experiment', None),
+        ('fake_hostname', 'ignore_experiment', 'fake_metric_name')]
     mock_options = mock.Mock()
     mock_options.rrddir_prefix = os.path.join(self._testdata_dir, 'rrd')
-    mock_options.ignored_experiments = []
+    mock_options.ignored_experiments = ['ignore_experiment']
     mock_options.pretty_json = None
     mock_options.compress = True
     mock_options.output = 'output'
     mock_options.ts_start = 0
     mock_options.ts_end = 10
-    expected_value = ('{"sample": [{"timestamp": 0, "value": 0.0}], '
-                      '"metric": "fake_metric", "hostname": "fake_hostname", '
-                      '"experiment": "fake_experiment"}\n')
+    expected_value = (
+        '{"sample": [{"timestamp": 0, "value": 0.0}], '
+        '"metric": "fake_metric_name", "hostname": "fake_hostname", '
+        '"experiment": "fake_experiment"}\n')
 
     mlab_export.rrd_export(mock_options)
     returned_value = str_fd.getvalue()
 
     self.assertEqual(returned_value, expected_value)
-    mock_get_canonical_names.assert_called_with(
-        'file1', 'skipped_metric', mock_options)
+    mock_get_canonical_names.assert_any_call('file1', 'value', mock_options)
+    mock_get_canonical_names.assert_any_call(
+        'file1', 'unknown_metric', mock_options)
+    mock_get_canonical_names.assert_any_call(
+        'file1', 'ignored_metric', mock_options)
     mock_open.assert_called_with('output.gz', 'w')
     self.assertTrue(mock_close.called)
+    self.assertTrue(mock_make_output_dirs.called)
 
   @mock.patch('mlab_export.default_output_name')
-  @mock.patch('mlab_export.assert_start_and_end_times')
   @mock.patch('mlab_export.default_start_time')
   @mock.patch('mlab_export.read_metric_map')
   def testunit_init_args(self, mock_read_metric_map,
                          mock_default_start_time,
-                         mock_assert_start_and_end_times,
                          mock_default_output_name):
     mock_options = disable_show_options(mock.Mock())
     mock_options.rrddir_prefix = os.path.join(self._testdata_dir, 'rrd')
--- a/export/mlab_export_test.py
+++ b/export/mlab_export_test.py
@@ -311,9 +311,7 @@ class MlabExport_GlobalTests(unittest.TestCase):
     expected_files = [rrd_dir + '/mlab2.nuq0t/file2.rrd',
                       rrd_dir + '/mlab2.nuq0t/file1.rrd']
 
-    returned_files = [ fname for fname in mlab_export.get_rrd_files(rrd_dir) ]
-
-    self.assertEqual(returned_files, expected_files)
+    self.assertEqual(mlab_export.get_rrd_files(rrd_dir), expected_files)
 
   @mock.patch('mlab_export.get_rrd_files')
   @mock.patch('mlab_export.rrdtool.fetch')
--- a/export/mlab_export_test.py
+++ b/export/mlab_export_test.py
@@ -13,10 +13,8 @@
 # limitations under the License.
 """Unit tests for collectd-mlab."""
 
-import errno
 import os
 import StringIO
-import time
 import unittest
 
 # Third-party modules.
--- a/export/mlab_export_test.py
+++ b/export/mlab_export_test.py
@@ -1,0 +1,492 @@
+# Copyright 2014 Google Inc. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Unit tests for collectd-mlab."""
+
+import os
+import errno
+import StringIO
+import time
+import unittest
+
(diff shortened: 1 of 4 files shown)


export/mlab_export_test.py
```diff
@@ -0,0 +1,492 @@
...
+class MlabExport_GlobalTests(unittest.TestCase):
+
+  def setUp(self):
+    self._testdata_dir = os.path.join(
+        os.path.dirname(mlab_export.__file__), 'testdata')
+
+  def testunit_LastTime(self):
+    fake_tstamp = os.path.join(self._testdata_dir, 'fake.tstamp')
+    # Artificially set fake timestamp to one hour ago.
+    one_hour_ago = int(time.time() - 3600)
```
mtlynch: I prefer to limit the amount of calls to the real system as much as possible, especially for something like time which we know will give us different results each time. Suggest hardcoding a timestamp (same for other `time.time()` calls in this file)

export/mlab_export_test.py
```diff
@@ -0,0 +1,492 @@
...
+      return_tstamp = last_time.get_mtime()
+
+    self.assertTrue(return_lock)
+    self.assertEqual(one_hour_ago, return_tstamp)
+
+  def testunit_LastTime_first_creation_RETURNS_zero(self):
+    fake_tstamp = os.path.join(self._testdata_dir, 'no_such_file.tstamp')
+    # Make sure that fake timestamp file actually doesn't exist.
+    if os.path.exists(fake_tstamp):
+      os.remove(fake_tstamp)
```
mtlynch: Does anything actually generate this file? If not, suggest removing 74-75, but I noticed that this file is checked into source control. Do we use it for anything except just immediately deleting it?

export/mlab_export_test.py
```diff
@@ -0,0 +1,492 @@
...
+    fake_ts = 1410341339
+    expected_6 = fake_ts - 5
+    expected_10 = fake_ts - 9
+    expected_30 = fake_ts - 29
+
+    ts_6 = mlab_export.align_timestamp(fake_ts, 6)
+    ts_10 = mlab_export.align_timestamp(fake_ts, 10)
+    ts_30 = mlab_export.align_timestamp(fake_ts, 30)
+
+    self.assertEqual(ts_6, expected_6)
```
mtlynch: This seems unnecessarily verbose. Could we just do:
```
self.assertEqual(mlab_export.align_timestamp(1410341339, 6), 1410341334)
self.assertEqual(mlab_export.align_timestamp(1410341339, 10), 1410341330)
self.assertEqual(mlab_export.align_timestamp(1410341339, 30), 1410341310)
```

Also, consider using numbers that are easy for the average person to verify in their head. Like align(15, 6) == 12 is very easy to verify without a calculator. It's not a realistic timestamp, but small vs. large numbers doesn't seem like the thing that's likely to affect the test.

export/mlab_export_test.py
```diff
@@ -0,0 +1,492 @@
...
+    self.assertEqual(ts_30, expected_30)
+
+  @mock.patch('mlab_export.time.time')
+  def testunit_default_end_time(self, mock_time):
+    mock_time.return_value = 1401401409
+    mock_options = mock.Mock()
+    mock_options.length = 3600
+    mock_options.step = 10
+    mock_options.update = False
+    expected_ts_end = 1401400800
```
mtlynch: Same note here. Can we use numbers that are easy to verify mentally? Like changing length to `5000`

export/mlab_export_test.py
```diff
@@ -0,0 +1,492 @@
...
+    mock_options.length = 3600
+    mock_options.ts_end = 1401401409 + mock_options.length
+    mock_options.ts_start = 1401401409
+
+    try:
+      mlab_export.assert_start_and_end_times(mock_options)
+    except mlab_export.TimeOptionError as err:
+      self.fail('Unexpected exception: %s' % err)
+
+    self.assertTrue(mock_logging.called)
```
mtlynch: Eh I feel like the logging is just an implementation detail and not part of the function contract so we shouldn't test this behavior.

export/mlab_export_test.py
```diff
@@ -0,0 +1,492 @@
...
+
+    returned_start = mlab_export.default_start_time(mock_options, ts_previous)
+
+    self.assertEqual(returned_start, expected_start)
+
+  @mock.patch('mlab_export.logging.debug')
+  def testunit_assert_start_and_end_times(self, mock_logging):
+    mock_options = mock.Mock()
+    mock_options.length = 3600
+    mock_options.ts_end = 1401401409 + mock_options.length
```
mtlynch: If we're doing arithmetic anyway, suggest reordering and making it `mock_options.ts_end = mock_options.ts_start + mock_options.length`

export/mlab_export_test.py
```diff
@@ -0,0 +1,492 @@
...
+                      mlab_export.assert_start_and_end_times, mock_options)
+
+  def testunit_assert_start_and_end_times_WHEN_too_short_RAISES_TimeOptionError(
+      self):
+    mock_options = mock.Mock()
+    mock_options.length = 3600
+    # Start & end should be at least 'length' apart. Subtract 1 to the
+    # difference too short.
+    mock_options.ts_end = 1401401409 + mock_options.length - 1
+    mock_options.ts_start = 1401401409
```
mtlynch: Same suggestion about making it `mock_options.ts_end = mock_options.ts_start + mock_options.length`

export/mlab_export_test.py
```diff
@@ -0,0 +1,492 @@
...
+        'disk-dm-0.disk_time.write': 'disk.swap.io.time.write'}
+
+    (returned_host, returned_experiment, returned_metric) = (
+        mlab_export.get_canonical_names(
+            fake_filename, value_name, mock_options))
+    
+    self.assertEqual(returned_host, expected_host)
+    self.assertEqual(returned_experiment, expected_experiment)
+    self.assertEqual(returned_metric, expected_metric)
+    self.assertTrue(mock_logging.called)
```
mtlynch: Same note about logging not part of the function contract. I'll stop commenting further instances of this.

export/mlab_export_test.py
```diff
@@ -0,0 +1,492 @@
...
+    returned_value = output.getvalue()
+ 
+    self.assertEqual(returned_value, expected_value)
+
+  def testunit_get_rrd_files(self):
+    rrd_dir = os.path.join(os.path.abspath(self._testdata_dir), 'rrd')
+    expected_files = [rrd_dir + '/mlab2.nuq0t/file2.rrd',
+                      rrd_dir + '/mlab2.nuq0t/file1.rrd']
+
+    returned_files = [ fname for fname in mlab_export.get_rrd_files(rrd_dir) ]
```
mtlynch: Style guidelines say no leading/trailing whitespace in a list comprehension.

mtlynch: I'm admittedly terrible at reading list comprehensions, but I can't figure out what this one does. How is it different from

```
returned_files = mlab_export.get_rrd_files(rrd_dir)
```

export/mlab_export_test.py
```diff
@@ -0,0 +1,492 @@
...
+    str_fd = StringIO.StringIO()
+    mock_open.return_value = str_fd
+    mock_get_rrd_files.return_value = ['file1']
+    mock_rrdtool_fetch.return_value = [
+        (0, 10, 10), ['value', 'skipped_metric'], [(0.0,0.0)]]
+    mock_get_canonical_names.side_effect = [
+        ('host', 'exp', 'metric'), ('host', 'exp', None)]
+    mock_options = mock.Mock()
+    mock_options.rrddir_prefix = os.path.join(self._testdata_dir, 'rrd')
+    mock_options.ignored_experiments = []
```
mtlynch: Can you add a test to exercise the `ignored_experiments` functionality?

export/mlab_export_test.py
```diff
@@ -0,0 +1,492 @@
...
+  def testunit_rrd_export(
+      self, mock_close, mock_open, mock_get_canonical_names, mock_rrdtool_fetch,
+      mock_get_rrd_files):
+    str_fd = StringIO.StringIO()
+    mock_open.return_value = str_fd
+    mock_get_rrd_files.return_value = ['file1']
+    mock_rrdtool_fetch.return_value = [
+        (0, 10, 10), ['value', 'skipped_metric'], [(0.0,0.0)]]
+    mock_get_canonical_names.side_effect = [
+        ('host', 'exp', 'metric'), ('host', 'exp', None)]
```
mtlynch: Can we use different names? For things like `metric` that's the name of the key in the JSON output, so it's tough to figure out looking at the code if an instance of `metric` is the key or the dummy value. Suggest using names that are more obviously not the keys, such as `foo_metric`, `dummy_host`, etc.

export/mlab_export_test.py
```diff
@@ -0,0 +1,492 @@
...
+    self.assertTrue(mock_read_metric_map.called)
+    self.assertTrue(mock_logging.called)
+    self.assertTrue(mock_default_start_time.called)
+    self.assertTrue(mock_default_end_time.called)
+    self.assertTrue(mock_assert_start_and_end_times.called)
+    self.assertTrue(mock_default_output_name.called)
+    self.assertEqual(mock_options.rrddir_prefix[-1], '/')
+
+  @mock.patch('mlab_export.os.makedirs')
+  def testunit_default_output_name(self, mock_makedirs):
```
mtlynch: minor: this test is out of order

export/mlab_export_test.py
```diff
@@ -0,0 +1,492 @@
...
+
+    mlab_export.init_args(mock_options, ts_previous)
+
+    self.assertTrue(mock_read_metric_map.called)
+    self.assertTrue(mock_logging.called)
+    self.assertTrue(mock_default_start_time.called)
+    self.assertTrue(mock_default_end_time.called)
+    self.assertTrue(mock_assert_start_and_end_times.called)
+    self.assertTrue(mock_default_output_name.called)
+    self.assertEqual(mock_options.rrddir_prefix[-1], '/')
```
mtlynch: I feel like we're not getting much assurance out of this test that values are being set correctly. Can we mock the returns and check that they make their way correctly into the `mock_options`?

mtlynch:
Finished this round of review.

export/mlab_export_test.py
stephen-soltesz: No problem. This change replaces uses of time.time() with a static timestamp.

This change also includes the minimum additional changes to bring the tests in sync with mlab_export.py. 7eb66eba03850226e3eacab9eb8c7def00735bf4

export/mlab_export_test.py
stephen-soltesz: You're right this file does not belong in the repository. Removed: 585717a63ee5c636a7e7fed8fe8495795a8449e6

export/mlab_export_test.py
stephen-soltesz: No problem. The test for align_timestamp is much shorter now: 1abb951eec275dd8c214cdebae24b0f8e62618ed

export/mlab_export_test.py
stephen-soltesz: default_end_time no longer exists, so this test is deleted. But, 7eb66eba03850226e3eacab9eb8c7def00735bf4 should help simplify the use of timestamps throughout.

export/mlab_export_test.py
stephen-soltesz: I agree. In this case mocking the logging method helped verify that the `assert_start_and_end_times` method ran to completion, but better to check an actual return value. So, I made this method return `True` on success. 3f368927e3d246f42f8f803a8f2b77cd226357ad

The other motivation for mocking the logging methods was to prevent the extra output when running the tests. But, just disabling the logging output is better than making it part of the tests. So, I removed all logging mocks in: f42298b13fc0d2fbefc585b3f85a86e1f44ec051

export/mlab_export_test.py
stephen-soltesz: No problem. This is consistent with using the static timestamp throughout. fc7be16fe64f5b901e72fc503c5c902407a551f1

export/mlab_export_test.py
stephen-soltesz: No problem. Also addressed in: 7eb66eba03850226e3eacab9eb8c7def00735bf4

export/mlab_export_test.py
stephen-soltesz: Good catch. The list comprehension is not necessary here. At one time I was experimenting with using `yield` in get_rrd_files, but that required actually calling the method N times, which the list comprehension did. But, after deciding that `yield` was not really helpful, I forgot to update the test. The test is simpler now: 438fc90cac2709ac217be9175f1a53b20feea305

export/mlab_export_test.py
stephen-soltesz: This is a good suggestion. I've added some extra setup for the rrd_export test to include the ignored_experiments feature. 44da703486401769d24ae7f3a3342621706f4c69

However, I would like to reconsider this test -- maybe split it into two parts; it's looking kind of complex.

export/mlab_export_test.py
stephen-soltesz: That's a good suggestion. The names are different from the keys: 7879a698521c4aba1066892690d2e96b4064fb60

export/mlab_export_test.py
stephen-soltesz: Ah. I moved this one, but think that I should revisit the ordering for other tests to match mlab_export as well. 6614be2e8010b484972001e41cda2601bbbb5d74

export/mlab_export_test.py
stephen-soltesz: That's a much better suggestion. The init_args test now checks the assigned values. e387443d2ff01604c6105c56983e9022e1caf1a1

stephen-soltesz:
Since having the mlab_export & mlab_export_test pull requests separated was not a helpful approach, I've merged this branch (with all changes) into pull https://github.com/m-lab/collectd-mlab/pull/7. So, we can continue from there with both files together and close this pull request.

export/mlab_export_test.py
mtlynch: Much easier to read, thanks.

PR #10, opened 2015-01-25
Fix mlab_test.py improper usage of assert_called_once.

stephen-soltesz:
This change corrects usage of assert_called_once to an assertion that
the mock was called at all. The method assert_called_once does not
actually exist. And, default mocks allow this behavior without error.

A future change should consider using one of the 'spec', or 'spec_set'
features on all mocks to enforce strict presence for any used
attributes.


mtlynch:
LGTM

PR #11, opened 2015-03-02
Add a nagios plugin and tests for check_collectd_mlab.py

stephen-soltesz:
This change includes check_collectd_mlab.py and associated tests.
check_collectd_mlab.py is a nagios plugin that checks the installation
and runtime health of collectd-mlab.

### DIFF
--- a/monitoring/check_collectd_mlab.py
+++ b/monitoring/check_collectd_mlab.py
@@ -1,0 +1,369 @@
+#!/usr/bin/python
+# Copyright 2014 Google Inc. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Summary:
+  check_collectd_mlab.py is a nagios plugin that checks the health of
+  collectd-mlab.
+
+  check_collectd_mlab.py should be installed and run form the host context.
+
+When check_collectd_mlab.py returns an OKAY status, the following are true:
+ * collectd is installed in the utility slice.
+ * collectd responds to commands sent over the collectd unix socket.
+ * collectd is running on a writable filesystem.
+ * vsys configuration is correct for the backend, frontend, and slice acl.
+ * resource usage by the collectd process and slice are normal.
+ * TODO: make resource usage checks configurable.
+ * TODO: check the last disk sync time recently and without too much error.
+
+If any of the above checks fail, check_collectd_mlab.py exits with a nagios
+error code and descriptive message.
+
+Installation:
+  /usr/lib/nagios/plugins/
+
+Example usage:
+  $ ./check_collectd_mlab.py
+  OKAY: Blue skies smiling at me: 1.27 sec
+
+If collectd is not running:
+  $ ./check_collectd_mlab.py
+  CRITICAL: collectd unixsock not present:
+       /vservers/mlab_utility/var/run/collectd-unixsock.
+"""
+import os
+import signal
+import socket
+import subprocess
+import sys
+import time
+
+
+# These values expect that:
+# * This script runs in host context.
+# * collectd runs in mlab_utility slice.
+# * collectd uses default pidfile name.
+# * collectd uses default unixsock name.
+SLICENAME = 'mlab_utility'
+COLLECTD_BIN = '/vservers/%s/usr/sbin/collectd' % SLICENAME
+COLLECTD_PID = '/vservers/%s/var/run/collectd.pid' % SLICENAME
+COLLECTD_UNIXSOCK = '/vservers/%s/var/run/collectd-unixsock' % SLICENAME
+
+HOSTNAME = socket.gethostname()
+VSYSPATH_ACL = '/vsys/vs_resource_backend.acl'
+VSYSPATH_BACKEND = '/vsys/vs_resource_backend'
+VSYSPATH_SLICE = '/vservers/%s/vsys/vs_resource_backend.in' % SLICENAME
+DEFAULT_TIMEOUT = 60
+
+# Canonical, nagios exit codes.
+STATE_OK = 0
+STATE_WARNING = 1
+STATE_CRITICAL = 2
+STATE_UNKNOWN = 3
+STATUS_MESSAGES = {
+    STATE_OK: 'OKAY:',
+    STATE_WARNING: 'WARNING:',
+    STATE_CRITICAL: 'CRITICAL:',
+    STATE_UNKNOWN: 'UNKNOWN:'
+}
+
+
+class Error(Exception):
+  """Base error class for this module."""
+
+
+class NagiosStateError(Error):
+  """A generic nagios status error."""
+  status_code = STATE_UNKNOWN
+  def __init__(self, message, status_code=None):
+    if status_code:
+      self.status_code = status_code
+    Error.__init__(self, message)
+
+
+class CriticalError(NagiosStateError):
+  """A critical error has occurred."""
+  status_code = STATE_CRITICAL
+
+
+class TimeoutError(Exception):
+  """A timeout has occurred."""
+
+
+def sock_connect(path):
+  """Creates a unix domain, stream socket and connects to path.
+
+  Args:
+    path: str, absolute path to unix socket name.
+  Returns:
+    AF_UNIX socket.socket of type SOCK_STREAM on success, None on error.
+  """
+  try:
+    sock = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
+    sock.connect(path)
+    return sock
+  except socket.error:
+    return None
+
+
+def sock_sendcmd(sock, command):
+  """Writes command to socket.
+
+  Args:
+    sock: socket.socket, connected unix domain socket.
+    command: str, the command to send over socket.
+  Returns:
+    int, positive on success, less than or equal to zero on error.
+  """
+  sock.send(command + '\n')
+  status_message = sock_readline(sock)
+  if not status_message:
+    return 0
+  code, _ = status_message.split(' ', 1)
+  try:
+    return int(code)
+  except ValueError:
+    return 0
+
+
+def sock_readline(sock):
+  """Reads one line from sock, until first newline character.
+
+  Args:
+     sock: socket.socket, to read line from this socket.
+  Returns:
+     str, the line read, or empty string on error.
+  """
+  try:
+    data = ''
+    buf = []
+    while data != '\n':
+      data = sock.recv(1)
+      if not data:
+        break
+      if data != '\n':
+        buf.append(data)
+    return ''.join(buf)
+  except socket.error:
+    return ''
+
+
+def assert_collectd_installed():
+  """Asserts that collectd is installed in the utility slice.
+
+  Raises:
+    CriticalError, if an error occurs.
+  """
+  # Is collectd installed?
+  if not os.path.exists(COLLECTD_BIN):
+    raise CriticalError('collectd binary not present: %s.' % COLLECTD_BIN)
+
+  # Is collectd socket available?
+  if not os.path.exists(COLLECTD_UNIXSOCK):
+    raise CriticalError(
+        'collectd unixsock not present: %s.' % COLLECTD_UNIXSOCK)
+
+
+def assert_collectd_responds():
+  """Asserts that collectd is responding over the COLLECTD_UNIXSOCK.
+
+  Raises:
+    CriticalError if an error occurs.
+  """
+
+  # Is filesystem read-write ok?
+  if not os.access(COLLECTD_PID, os.W_OK):
+    raise CriticalError('collectd filesystem is NOT_WRITEABLE!')
+
+  # Is collectd responsive over unix socket?
+  sock = sock_connect(COLLECTD_UNIXSOCK)
+  if sock is None:
+    raise CriticalError(
+        'CONNECT_FAILED but unixsock present (%s)!' % COLLECTD_UNIXSOCK)
+
+  # Can we request a value over socket?
+  val = sock_sendcmd(sock, 'GETVAL "%s/meta/timer-read"' % HOSTNAME)
+  if val <= 0:
+    raise CriticalError(
+        'SENDCMD_FAILED but collectd unixsock is open.')
+
+  # Read as many lines as reported back from command.
+  for _ in xrange(0, val):
+    if not sock_readline(sock):
+      raise CriticalError('EMPTY_MESSAGE read from collectd socket.')
+
+
+def assert_collectd_vsys_setup():
+  """Asserts that vsys configuration is complete for mlab_utility.
+
+  Raises:
+    CriticalError if an error occurs.
+  """
+  # Is the vsys backend script installed?
+  if not os.path.exists(VSYSPATH_BACKEND):
+    raise CriticalError('NO_VSYS_BACKEND %s is missing!' % VSYSPATH_BACKEND)
+
+  # Is the vsys frontend FIFO in the slice context?
+  if not os.path.exists(VSYSPATH_SLICE):
+    raise CriticalError('NO_VSYS_SLICE %s is missing in slice' % VSYSPATH_SLICE)
+
+  # Is mlab_utility in the vsys acl for the backend script?
+  try:
+    acl = open(VSYSPATH_ACL, 'r').read().strip()
+  except IOError:
+    raise CriticalError('NO_VSYS_ACL cannot read vsys acl: %s' % VSYSPATH_ACL)
+
+  if SLICENAME not in acl:
+    raise CriticalError(
+        'NO_SLICE_IN_ACL %s is missing from ACL %s' % (SLICENAME, VSYSPATH_ACL))
+
+
+def run_collectd_nagios(host, metric, value, warning, critical):
+  """Runs collectd-nagios using given parameters as arguments.
+
+  Please see the collectd-nagios man page for documentation on the format of
+  'warning' and 'critical' thresholds.
+
+  Args:
+    host: str, experiment hostname.
+    metric: str, raw metric path. (e.g. network/if_octets-ipv6)
+    value: str, name of value within metric. (e.g. 'value', 'rx')
+    warning: str, a collectd-nagios warning threshold. (e.g. '0:20')
+    critical: str, a collectd-nagios critical threshold. (e.g. '0:30')
+  Returns:
+    exit code from collectd-nagios. Because this is a nagios plugin, these are
+        valid nagios exit states.
+  """
+  cmd = ('collectd-nagios -s {unixsock} -H {host} '+
+         '-n {metric} -d {value} -w {warning} '+
+         '-c {critical}')
+  cmd = cmd.format(unixsock=COLLECTD_UNIXSOCK, host=host,
+                   metric=metric, value=value, warning=warning,
+                   critical=critical)
+  child = subprocess.Popen(
+      cmd.split(), stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+  return child.wait()
+
+
+def assert_collectd_nagios_levels():
+  """Asserts actual values from collectd using collectd-nagios.
+
+  Asserts that storage quota is sufficient, metric collection is fast enough,
+  that CPU load is low, and memory usage is reasonable.
+
+  Raises:
+    NagiosStateError, if an error occurs.
+  """
+  # TODO: Make warning & critical thresholds configurable.
+
+  # Is utility slice quota ok?
+  exit_code = run_collectd_nagios(
+      'utility.mlab.'+HOSTNAME, 'storage/vs_quota_bytes-quota',
+      'used', '0:8000000000', '0:9000000000')
+  if exit_code != 0:
+    raise NagiosStateError('Storage quota usage is too high',
+                           status_code=exit_code)
+
+  # Is collectd cpu usage too high?
+  exit_code = run_collectd_nagios(
+      HOSTNAME, 'meta-collectd/process_cpu-system',
+      'value', '0:5', '0:10')
+  if exit_code != 0:
+    raise NagiosStateError('Collectd CPU usage is too high',
+                           status_code=exit_code)
+
+  # Actual memory usage should be pretty small, though vm size may be larger.
+  exit_code = run_collectd_nagios(
+      HOSTNAME, 'meta-collectd/process_memory-rss',
+      'value', '0:10000000', '0:15000000')
+  if exit_code != 0:
+    raise NagiosStateError('Collectd RSS memory usage is too high',
+                           status_code=exit_code)
+
+  # Is meta timer too high?
+  exit_code = run_collectd_nagios(
+      HOSTNAME, 'meta/timer-read',
+      'value', '0:6', '0:100')
+  if exit_code != 0:
+    raise NagiosStateError('Collectd mlab plugin taking too long to run',
+                           status_code=exit_code)
+
+
+def assert_disk_last_sync_time():
+  """Asserts the last sync time for vserver disk limits."""
+  # TODO: When was vserver quota last sync'd?
+  # TODO: How do we want to check this?
+  pass
+
+
+def check_collectd():
+  """Checks environment for signs of normal collectd-mlab behavior.
+  
+  Returns:
+    Tuple of (int, str), corresponding to the (nagios_state, error_message).
+  """
+  t_start = time.time()
+
+  # Check all critical conditions first.
+  try:
+    assert_collectd_installed()
+    assert_collectd_responds()
+    assert_collectd_vsys_setup()
+    assert_disk_last_sync_time()
+  except CriticalError as err:
+    return (err.status_code, str(err))
+
+  # Since the above establishes that collectd is working, now check collectd
+  # resource usage thresholds *from* collectd.
+  try:
+    assert_collectd_nagios_levels()
+  except NagiosStateError as err:
+    return (err.status_code, str(err))
+
+  # We've made it this far, so everything appears to be ok.
+  msg = 'Blue skies smiling at me: %0.2f sec' % (time.time() - t_start)
+  return (STATE_OK, msg)
+
+
+def init_alarm(timeout):
+  """Assigns a SIGALARM handler for timeout seconds."""
+  def handler(unused_signum, unused_frame):
+    """Raise a TimeoutError when called"""
+    raise TimeoutError('Timeout after %s' % timeout)
+  signal.signal(signal.SIGALRM, handler)
+  signal.alarm(timeout)
+
+
+def cancel_alarm():
+  """Cancels pending alarm."""
+  signal.alarm(0)
+
+
+def main():
+  init_alarm(DEFAULT_TIMEOUT)
+
+  try:
+    (status_code, message) = check_collectd()
+  except Exception as err:  # pylint: disable=W0703
+    status_code = STATE_UNKNOWN
+    message = str(err)
+
+  cancel_alarm()
+
+  print STATUS_MESSAGES.get(status_code, 'UNKNOWN:'), message
+  sys.exit(status_code)
+
+
+if __name__ == "__main__":
+  main()  # pragma: no cover.
--- a/monitoring/check_collectd_mlab.py
+++ b/monitoring/check_collectd_mlab.py
@@ -159,32 +159,45 @@ def sock_connect(path):
   Args:
     path: str, absolute path to unix socket name.
   Returns:
-    AF_UNIX socket.socket of type SOCK_STREAM on success, None on error.
+    AF_UNIX socket.socket of type SOCK_STREAM.
+  Raises:
+    CriticalError, if socket connection fails.
   """
   try:
     sock = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
     sock.connect(path)
     return sock
-  except socket.error:
-    return None
+  except socket.error as err:
+    raise SocketConnectionCriticalError(
+        'Failed to connect to socket %s! Received socket.error: %s' %
+        (path, err))
 
 
 def sock_sendcmd(sock, command):
-  """Writes command to socket.
+  """Writes command to collectd UnixSock socket.
+
+  For more information, see also the collectd UnixSock socket protocol:
+    https://collectd.org/wiki/index.php/Plain_text_protocol
 
   Args:
     sock: socket.socket, connected unix domain socket.
     command: str, the command to send over socket.
   Returns:
-    int, positive on success, less than or equal to zero on error.
+    int, the status code returned by collectd. The value is positive on
+        success, less than or equal to zero on error.
+  Raises:
+    CriticalError, if socket send fails.
   """
-  sock.send(command + '\n')
-  status_message = sock_readline(sock)
-  if not status_message:
-    return 0
-  code, _ = status_message.split(' ', 1)
   try:
-    return int(code)
+    sock.send(command + '\n')
+  except socket.error as err:
+    raise SocketSendCommandCriticalError(
+        'Sending %s failed: %s' % (command, err))
+
+  status_message = sock_readline(sock)
+  code = status_message.split(' ', 1)
+  try:
+    return int(code[0])
   except ValueError:
     return 0
 
@@ -196,19 +209,19 @@ def sock_readline(sock):
      sock: socket.socket, to read line from this socket.
   Returns:
      str, the line read, or empty string on error.
+  Raises:
+    CriticalError, if socket read fails.
   """
   try:
-    data = ''
     buf = []
-    while data != '\n':
+    data = sock.recv(1)
+    while data and data != '\n':
+      buf.append(data)
       data = sock.recv(1)
-      if not data:
-        break
-      if data != '\n':
-        buf.append(data)
     return ''.join(buf)
-  except socket.error:
-    return ''
+  except socket.error as err:
+    raise SocketReadlineCriticalError(
+        'Failed to read message from collectd. Received error: %s' % err)
 
 
 def assert_collectd_installed():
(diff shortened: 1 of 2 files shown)


stephen-soltesz:
I see the review is paused. Would it be disruptive if I began making changes based on the comments so far?

stephen-soltesz:
PTAL?

monitoring/check_collectd_mlab.py
```diff
@@ -0,0 +1,461 @@
...
+  """A generic nagios status error."""
+
+  def __init__(self, message, status_code=STATE_UNKNOWN):
+    self.status_code = status_code
+    Error.__init__(self, message)
+
+
+class CriticalError(Error):
+  """A base class for all critical errors."""
+  pass
```
mtlynch: I think this error scheme is much easier to understand.

Two notes:
* I suggest not putting "Critical" in the names because it tightly ties the error to how the error is interpreted. If we have something that's a critical error now, but we later decide it's not critical, it would be nice if that only required changing its parent class rather than its name and all references to it.
* I think the Error classes themselves should do more of the heavy lifting wrt messages. For example, to raise a `MissingSliceFromVsysAclCriticalError` the caller is doing this:

```
raise MissingSliceFromVsysAclCriticalError(
        'Slice name %s is missing from ACL: %s' % (SLICENAME, VSYSPATH_ACL))
```

Whereas I think I would prefer to see it like this:

```
class MissingSliceFromVsysAclCriticalError(CriticalError):
  def __init__(self, slice_name, vsyspath_acl):
    message = 'Slice name %s is missing from ACL: %s' % (slice_name, vsyspath_acl))
    CriticalError.__init__(self, message)

...
raise MissingSliceFromVsysAclCriticalError(SLICENAME, VSYSPATH_ACL)
```

monitoring/check_collectd_mlab_test.py
```diff
@@ -0,0 +1,379 @@
...
+        # Send "number of values" response.
+        if self._failure_mode == 'send_bad_reply':
+          client.sendall('not-a-number junk\n')
+        elif self._failure_mode == 'send_empty_reply':
+          client.sendall('\n')
+        else:
+          client.sendall('1 default reply\n')
+
+        # Send "actual values" response.
+        if self._failure_mode == 'close_after_reply':
```
mtlynch: Suggest renaming to `close_during_reply` because it took me a bit looking at 101-108 to figure out why we were saying we were closing _after_ reply when we seemed to be closing _before_ the reply.

monitoring/check_collectd_mlab_test.py
```diff
@@ -0,0 +1,379 @@
...
+          return
+        try:
+          client.sendall('default=1.0\n')
+        except socket.error:
+          # Incase of sigpipe due to client closing connection first.
+          pass
+    # pylint: enable=E1101
+
+
+class MlabNagiosCollectdTests(unittest.TestCase):
```
mtlynch: Should be `MLab`

monitoring/check_collectd_mlab_test.py
```diff
@@ -0,0 +1,379 @@
...
+    check_collectd_mlab.COLLECTD_PID = (
+        os.path.join(self._testdata_dir, 'fake_pid'))
+    check_collectd_mlab.COLLECTD_UNIXSOCK = self.fake_socket
+
+    self.assertRaises(
+        check_collectd_mlab.SocketReadlineCriticalError,
+        check_collectd_mlab.assert_collectd_responds)
+
+
+class MlabNagiosTests(unittest.TestCase):
```
mtlynch: `MLab`*

mtlynch:
Sorry I didn't see your questions until now. I generally don't read through the Github notifications until I set time aside to dive into the code review again, so if you ever need my attention on something sooner than when I'm next going to pick up the review, please shoot me an email.

monitoring/check_collectd_mlab_test.py
```diff
@@ -0,0 +1,379 @@
...
+  def setUp(self):
+    self._testdata_dir = os.path.join(
+        os.path.dirname(check_collectd_mlab.__file__), 'testdata')
+
+  def testunit_sock_connect_WHEN_no_socket_RAISES_CriticalError(self):
+    self.assertRaises(
+        check_collectd_mlab.SocketConnectionCriticalError,
+        check_collectd_mlab.sock_connect, 'no_socket_name')
+
+  def testunit_sock_readline_WHEN_socket_error_RETURNS_empty_string(self):
```
mtlynch: Test name doesn't seem to match behavior here because it's raising an error instead of returning empty string.

monitoring/check_collectd_mlab.py
```diff
@@ -0,0 +1,461 @@
...
+  """
+  # TODO: Make warning & critical thresholds configurable.
+
+  # Is utility slice quota ok?
+  exit_code = run_collectd_nagios(
+      'utility.mlab.' + HOSTNAME, 'storage/vs_quota_bytes-quota',
+      'used', _mb_to_bytes(8000), _mb_to_bytes(9000))
+  if exit_code != 0:
+    raise NagiosStateError('Storage quota usage is too high',
+                           status_code=exit_code)
```
mtlynch: Why use keyword arguments here? I feel like 

```
raise NagiosStateError('Storage quota usage is too high', exit_code)
```
is pretty clear (and has the benefit of fitting on one line)

monitoring/check_collectd_mlab.py
```diff
@@ -0,0 +1,461 @@
...
+def main():
+  init_alarm(DEFAULT_TIMEOUT)
+
+  try:
+    (status_code, message) = check_collectd()
+  except Exception as err:  # pylint: disable=W0703
+    status_code = STATE_UNKNOWN
+    message = str(err)
+
+  cancel_alarm()
```
mtlynch: (optional) Given my perennial love for context handlers, I suggest doing this in a context handler.

monitoring/check_collectd_mlab_test.py
```diff
@@ -0,0 +1,379 @@
...
+    self.assertRaises(
+        check_collectd_mlab.MissingBinaryCriticalError,
+        check_collectd_mlab.assert_collectd_installed)
+
+  def testunit_assert_collectd_installed_WHEN_nagios_bin_missing_CriticalError(
+      self):
+    check_collectd_mlab.COLLECTD_BIN = (
+        os.path.join(self._testdata_dir, 'fake_bin'))
+    check_collectd_mlab.COLLECTD_NAGIOS = (
+        os.path.join(self._testdata_dir, 'does_not_exist'))
```
mtlynch: I think this is a good reason to put everything in `setUp`. It's not super obvious when reading it that fake_bin is supposed to exist and does_not_exist doesn't. Reading through it just now, I got confused because I thought since it's a test for missing files, neither file existed. If `setUp` included:

```
check_collectd_mlab.COLLECTD_BIN = os.path.join(self._testdata_dir, 'fake_bin')
check_collectd_mlab.COLLECTD_NAGIOS = os.path.join(self._testdata_dir, 'fake_bin')
check_collectd_mlab.COLLECTD_UNIXSOCK = 'fake_sock'
check_collectd_mlab.VSYSPATH_BACKEND = os.path.join(self._testdata_dir, 'fake_backend')
check_collectd_mlab.VSYSPATH_SLICE = os.path.join(self._testdata_dir, 'fake_slice')
 check_collectd_mlab.VSYSPATH_ACL = os.path.join(self._testdata_dir, 'fake_acl')
```

Then when you set things to `does_not_exist` it's obvious that they're the only component that doesn't exist.

monitoring/check_collectd_mlab_test.py
```diff
@@ -0,0 +1,379 @@
...
+    check_collectd_mlab.VSYSPATH_BACKEND = os.path.join(
+        self._testdata_dir, 'fake_backend')
+    check_collectd_mlab.VSYSPATH_SLICE = os.path.join(
+        self._testdata_dir, 'fake_slice')
+    check_collectd_mlab.VSYSPATH_ACL = os.path.join(
+        self._testdata_dir, 'fake_acl')
+
+    self.assertRaises(
+        check_collectd_mlab.MissingSliceFromVsysAclCriticalError,
+        check_collectd_mlab.assert_collectd_vsys_setup)
```
mtlynch: This is a bit hard to understand. Is the test supposed to fail because the ACL doesn't include the slice name? It's confusing because `fake_backend` is supposed to be good, `fake_slice` is supposed to be good, so it's unintuitive that `fake_acl` is bad. Consider naming it something like `acl_missing_slice_name`.

monitoring/check_collectd_mlab_test.py
```diff
@@ -0,0 +1,379 @@
...
+    mock_run_collectd_nagios.side_effect = [0, 0, 1]
+    self.assertRaises(check_collectd_mlab.NagiosStateError,
+                      check_collectd_mlab.assert_collectd_nagios_levels)
+
+    mock_run_collectd_nagios.side_effect = [0, 0, 0, 1]
+    self.assertRaises(check_collectd_mlab.NagiosStateError,
+                      check_collectd_mlab.assert_collectd_nagios_levels)
+
+  def teststub_assert_disk_last_sync_time(self):
+    check_collectd_mlab.assert_disk_last_sync_time()
```
mtlynch: Eh I don't think there's much value to adding this.

monitoring/check_collectd_mlab_test.py
```diff
@@ -0,0 +1,379 @@
...
+    returned_value = check_collectd_mlab.run_collectd_nagios(
+        'host', 'metric', 'value', 'warning', 'critical')
+
+    self.assertEqual(returned_value, 2)
+    self.assertTrue(mock_popen.called)
+
+  @mock.patch('check_collectd_mlab.run_collectd_nagios')
+  def testcover_assert_collectd_nagios_levels(self, mock_run_collectd_nagios):
+    # This is not ideal. But, it's just a coverage test.
+    # Non-zero values cause a failue. Cause each call to to fail in sequence.
```
mtlynch: This makes me nervous because we lose coverage if someone adds a call to `run_collectd_nagios`. Could we base the return code on the metric parameter?

monitoring/check_collectd_mlab_test.py
```diff
@@ -0,0 +1,379 @@
...
+
+  @mock.patch('check_collectd_mlab.assert_collectd_installed')
+  @mock.patch('check_collectd_mlab.assert_collectd_responds')
+  @mock.patch('check_collectd_mlab.assert_collectd_vsys_setup')
+  @mock.patch('check_collectd_mlab.assert_collectd_nagios_levels')
+  @mock.patch('check_collectd_mlab.assert_disk_last_sync_time')
+  def testcover_check_collectd(
+      self, mock_last_sync_time, mock_collectd_nagios_levels, mock_vsys_setup,
+      mock_collectd_responds, mock_collectd_installed):
+
```
mtlynch: Extraneous newline

monitoring/check_collectd_mlab_test.py
```diff
@@ -0,0 +1,379 @@
...
+    mock_popen.return_value.wait.return_value = 2
+
+    returned_value = check_collectd_mlab.run_collectd_nagios(
+        'host', 'metric', 'value', 'warning', 'critical')
+
+    self.assertEqual(returned_value, 2)
+    self.assertTrue(mock_popen.called)
+
+  @mock.patch('check_collectd_mlab.run_collectd_nagios')
+  def testcover_assert_collectd_nagios_levels(self, mock_run_collectd_nagios):
```
mtlynch: What is a coverage test? I've never seen this before.

mtlynch:
Finished this round of review.

monitoring/check_collectd_mlab_test.py
stephen-soltesz: Right, the test should fail b/c the mlab_utility slice is not listed in the acl. I agree that the original name was confusing based on how other "fake_*" names are used.

New name is "acl_missing_slice_name".

monitoring/check_collectd_mlab.py
stephen-soltesz: Change ab8fbeb4b6260efa0c9557228cad2523fb2108a3 removes "Critical" from the subclasses of CriticalError and updates how error messages are formatted at the call sites.

I was not sure how common a pattern parameterized exceptions are for python. And, I worry about the extra boilerplate needed to support parameterized messages for single-use exception types. 

Still to clean up the call sites, I changed the CriticalError class to accept multiple arguments which are used to format the message. This ends up looking a little like the way logging methods are called. So, this:

        raise MissingSliceFromVsysAclCriticalError(
                'Slice name %s is missing from ACL: %s' % (SLICENAME, VSYSPATH_ACL))

changes to this:

        raise MissingSliceFromVsysAclCriticalError(
                'Slice name %s is missing from ACL: %s', SLICENAME, VSYSPATH_ACL)

Let me know what you think.

monitoring/check_collectd_mlab_test.py
stephen-soltesz: Fixed: 9189b1b84b7ed1a04cc81d9661bdd4a1400b4354

monitoring/check_collectd_mlab.py
stephen-soltesz: No problem. Included in the exception updates: ab8fbeb4b6260efa0c9557228cad2523fb2108a3

monitoring/check_collectd_mlab.py
stephen-soltesz: I gave it a shot. How does it look? f8fb71b97dc53c2976ac568e9dd69412365770e4

monitoring/check_collectd_mlab_test.py
stephen-soltesz: I agree. This works really well. Included in the test-rearrangement change: 97c0f41e85379ec201596246866aec425307a77c

monitoring/check_collectd_mlab_test.py
stephen-soltesz: Removed.

monitoring/check_collectd_mlab_test.py
stephen-soltesz: Removed. 37a53c89ed1d1fe056b3cc290a8812f3d6f6ca54

monitoring/check_collectd_mlab_test.py
stephen-soltesz: I made up the name. This is probably the lowliest of tests. It makes sure that the interpreter visits every line even when no functionality in particular is being tested. The best (not necessarily good) defense for this approach is that since python doesn't do static type checking, "silly" errors may not show up until run time (a variable is used before being defined, or a typo makes a variable appear to not exist). These tests give "silly" mistakes a chance to show up beforehand.

monitoring/check_collectd_mlab_test.py
stephen-soltesz: I'm afraid I don't understand your suggestion. Can you phrase it differently?

stephen-soltesz:
This took longer to get back to than I wanted. But, I think this is ready for another pass. PTAL? Thank you.

monitoring/check_collectd_mlab.py
mtlynch: Looks good!

monitoring/check_collectd_mlab_test.py
```diff
@@ -0,0 +1,301 @@
...
+
+class MLabNagiosSocketTests(unittest.TestCase):
+
+  def setUp(self):
+    self._testdata_dir = os.path.join(
+        os.path.dirname(check_collectd_mlab.__file__), 'testdata')
+
+  def testunit_sock_sendcmd_RETURNS_successfully(self):
+    mock_sock = mock.Mock(spec_set=socket.socket)
+    mock_sock.recv.side_effect = list('1 default reply\n')
```
mtlynch: Shouldn't this be `.return_value` in cases where it's not raising an exception?

monitoring/check_collectd_mlab_test.py
```diff
@@ -0,0 +1,301 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Unit tests for check_collectd_mlab check."""
+
+import os
+import socket
+import time
+import unittest2 as unittest
```
mtlynch: Why not just `import unittest`? Do we need to run this anywhere < 2.7?

monitoring/check_collectd_mlab_test.py
```diff
@@ -0,0 +1,301 @@
...
+    mock_sock.recv.side_effect = list('1 default reply\n')
+
+    returned_value = check_collectd_mlab.sock_sendcmd(
+        mock_sock, 'GETVAL "whatever"')
+
+    self.assertEqual(returned_value, 1)
+    mock_sock.send.assert_called_with('GETVAL "whatever"\n')
+
+  def testunit_sock_sendcmd_WHEN_receive_bad_reply_RETURNS_zero(self):
+    mock_sock = mock.Mock(spec_set=socket.socket)
```
mtlynch: Can we make this `self.mock_sock` and initialize it once in `setUp`?

monitoring/check_collectd_mlab_test.py
```diff
@@ -0,0 +1,301 @@
...
+import check_collectd_mlab
+
+# R0904: Too many public methods. Hard to avoid for unit tests.
+# pylint: disable=R0904
+
+
+class MLabNagiosSocketTests(unittest.TestCase):
+
+  def setUp(self):
+    self._testdata_dir = os.path.join(
```
mtlynch: minor, subjective, optional: I don't think there's much value in making test class attributes private because they don't have any clients trying to call them, so it unnecessarily adds typing.

monitoring/check_collectd_mlab_test.py
```diff
@@ -0,0 +1,301 @@
...
+    returned_value = check_collectd_mlab.sock_sendcmd(
+        mock_sock, 'GETVAL "whatever"')
+
+    self.assertEqual(returned_value, 0)
+    mock_sock.send.assert_called_with('GETVAL "whatever"\n')
+
+  def testunit_sock_connect_WHEN_no_socket_RAISES_Error(self):
+    self.assertRaises(
+        check_collectd_mlab.SocketConnectionError,
+        check_collectd_mlab.sock_connect, 'no_socket_name')
```
mtlynch: (subjective, optional) I just learned a new way of calling `assertRaises` that I view as more readable, but it's up to you if you prefer it to the usual way.

```
with self.assertRaises(check_collectd_mlab.SocketConnectionError):
  check_collectd_mlab.sock_connect('no_socket_name')
```

monitoring/check_collectd_mlab_test.py
```diff
@@ -0,0 +1,301 @@
...
+    mock_sock = mock.Mock(spec_set=socket.socket)
+    mock_sock.recv.side_effect = list('not-a-number junk\n')
+
+    returned_value = check_collectd_mlab.sock_sendcmd(
+        mock_sock, 'GETVAL "whatever"')
+
+    self.assertEqual(returned_value, 0)
+    mock_sock.send.assert_called_with('GETVAL "whatever"\n')
+
+  def testunit_sock_connect_WHEN_no_socket_RAISES_Error(self):
```
mtlynch: Why is `Error` capitalized here and in 63?

monitoring/check_collectd_mlab_test.py
```diff
@@ -0,0 +1,301 @@
...
+    self.assertRaises(
+        check_collectd_mlab.SocketConnectionError,
```
mtlynch: Suggest `invalid_socket_name` because we're passing a name, it's just invalid, right?

monitoring/check_collectd_mlab_test.py
```diff
@@ -0,0 +1,301 @@
...
+
+class MLabCollectdAssertionTests(unittest.TestCase):
+
+  def setUp(self):
+    self._testdata_dir = os.path.join(
+        os.path.dirname(check_collectd_mlab.__file__), 'testdata')
+    check_collectd_mlab.COLLECTD_BIN = (
+        os.path.join(self._testdata_dir, 'fake_bin'))
+    check_collectd_mlab.COLLECTD_NAGIOS = (
+        os.path.join(self._testdata_dir, 'fake_bin'))
```
mtlynch: Is it intentional that `COLLECTD_NAGIOS` and `COLLECTD_BIN` have the same name?

monitoring/check_collectd_mlab_test.py
```diff
@@ -0,0 +1,301 @@
...
+        os.path.join(self._testdata_dir, 'fake_pid'))
+    check_collectd_mlab.COLLECTD_UNIXSOCK = (
+        os.path.join(self._testdata_dir, 'fake_socket'))
+    check_collectd_mlab.VSYSPATH_BACKEND = (
+        os.path.join(self._testdata_dir, 'fake_backend'))
+    check_collectd_mlab.VSYSPATH_SLICE = (
+        os.path.join(self._testdata_dir, 'fake_slice'))
+
+  @mock.patch('check_collectd_mlab.sock_connect')
+  def testunit_assert_collectd_responds_WHEN_sock_sendcmd_fails(
```
mtlynch: "responds" is a bit unintuitive to me. What's responding? I'd expect `raises_error` or something.

monitoring/check_collectd_mlab_test.py
```diff
@@ -0,0 +1,301 @@
...
+        check_collectd_mlab.assert_collectd_responds)
+    mock_sock_connect.assert_called_with(
+        check_collectd_mlab.COLLECTD_UNIXSOCK)
+
+  @mock.patch('check_collectd_mlab.sock_connect')
+  def testunit_assert_collectd_responds_WHEN_sock_readline_fails(
+      self, mock_sock_connect):
+    mock_sock = mock.Mock(spec_set=socket.socket)
+    # After sending a default reply, send EOF ('') prematurely.
+    mock_sock.recv.side_effect = list('1 default reply\n') + ['']
```
mtlynch:
```
$ python -c "print (list('1 default reply\n') + [''])"
['1', ' ', 'd', 'e', 'f', 'a', 'u', 'l', 't', ' ', 'r', 'e', 'p', 'l', 'y', '\n', '']
```

I assume what we want is `['1 default reply\n', '']`

monitoring/check_collectd_mlab_test.py
```diff
@@ -0,0 +1,301 @@
...
+    # Mock out os.access to guarantee that write access is rejected.
+    mock_access.return_value = False
+
+    self.assertRaises(
+        check_collectd_mlab.ReadonlyFilesystemError,
+        check_collectd_mlab.assert_collectd_responds)
+    self.assertTrue(mock_access.called)
+
+  def testunit_assert_collectd_responds_WHEN_sock_connect_fails(self):
+    check_collectd_mlab.COLLECTD_UNIXSOCK = 'does_not_exist'
```
mtlynch: How does this differ from the test on 91?

monitoring/check_collectd_mlab_test.py
mtlynch: Hmm, now I have to figure out what I meant too...

Okay, so I was saying that this test is somewhat flaky because it depends on the exact number of times we call `run_collectd_nagios`, so I was thinking about a mock function that would check the metric (e.g. `meta/timer-read`) and return 1 for that specific call, 0 otherwise, but that's much more complicated and still leaves us with a pretty equally flaky test. Rescinded!

monitoring/check_collectd_mlab_test.py
```diff
@@ -0,0 +1,301 @@
...
+        self.fail('Alarm did not trigger.')  # pragma: no cover.
+      except check_collectd_mlab.TimeoutError:
+        pass
+
+  @mock.patch('sys.stdout')
+  @mock.patch('check_collectd_mlab.check_collectd')
+  def testcover_main(self, mock_check_status, mock_stdout):
+    mock_check_status.return_value = (check_collectd_mlab.STATE_OK, 'ok')
+    self.assertRaises(SystemExit, check_collectd_mlab.main)
+    self.assertTrue(len(mock_stdout.mock_calls) > 0)
```
mtlynch: Suggest `self.assertTrue(mock_stdout.called)`, ditto for 296

monitoring/check_collectd_mlab.py
mtlynch: Still seems a little strange to me, but from my brief search around, I think you're right in that Python isn't so big on baking in the error message the way I was talking about

mtlynch:
Finished this round of review

monitoring/check_collectd_mlab_test.py
stephen-soltesz: I guess this is a unittest-ism. A value assigned to `.return_value` is returned as-is when the mock is called.
```
>>> import mock
>>> m = mock.Mock()
>>> m.return_value = ['a', 'b', 'c']
>>> m()
['a', 'b', 'c']
>>> m()
['a', 'b', 'c']
```
Whereas the `.side_effect` attribute is a way of have different return values on subsequent calls to the same mock.
```
>>> m = mock.Mock()
>>> m.side_effect = ['a', 'b', 'c']
>>> m()
'a'
>>> m()
'b'
>>> m()
'c'
```

monitoring/check_collectd_mlab_test.py
stephen-soltesz: This sounds familiar, but it's been a while, and revisiting choices can be helpful in case circumstances have changed.

My thinking is that having the production environment be the same as the test-environment is advantageous. And, the current M-Lab node OS (CentOS 6) uses python 2.6.6.

monitoring/check_collectd_mlab_test.py
stephen-soltesz: Yes, that works. b1b85e5ebbd4024e8feff1045330e53a18433e34

monitoring/check_collectd_mlab_test.py
stephen-soltesz: That's a good point. I don't actually remember why I used the private prefix. And, it turns out I didn't need two of them! 5a760825102ce46a802d35a0508dd37c2d4b9a00

monitoring/check_collectd_mlab_test.py
stephen-soltesz: I like this too. Looking back I started with the "single line" form b/c the python 2.6 version of unittest does not support the context manager form. Since we're using unittest2 now, this is no problem. 4d8c888e7963df94c90db2eca80c201a55d0e2b3

I added a few newlines to help preserve the separation (where I thought it helped) between the "arrange, act, and assert" blocks of the test. Let me know if it looks alright.

monitoring/check_collectd_mlab_test.py
stephen-soltesz: It looks like the original form was "_RAISES_CriticalError" and when I generalized the exception names to not reference "Critical" these became "_RAISES_Error". Originally, the name / capitalization referenced an actual type so that seemed to make sense, rather than un-capitalizing it. But, since Error doesn't refer to an actual type now, I changed the capitalization to lowercase.  4d04ea89f6aaf748520be3f25ecc45dbd54e8162

monitoring/check_collectd_mlab_test.py
stephen-soltesz: That's right; it's just invalid. Sounds good. 5dc58156aa0c5cf646de49e2a8f6228a458d8aab

monitoring/check_collectd_mlab_test.py
stephen-soltesz: Yes, this was intentional. They just needed to point to something so presence checks passed but they need not be distinct. I'm open to adding a new file and clarifying the names: maybe "fake_collectd_bin" and "fake_nagios_bin".

monitoring/check_collectd_mlab_test.py
stephen-soltesz: I'm afraid I don't understand. Is the original method name unclear and/or the test name? The method under test is `assert_collectd_responds()`.

Maybe related, it looks like I use the suffix "_RAISES_error" inconsistently. In some cases this couldn't be added within 80 columns. :-/  I'm inclined to remove "_RAISES_error" suffix for the tests in the `MLabCollectdAssertionTests` class.

monitoring/check_collectd_mlab_test.py
stephen-soltesz: That's right. The "side_effect" values correspond to each value returned by `sock.recv(1)` in `sock_readline()`.

monitoring/check_collectd_mlab_test.py
stephen-soltesz: This one, `testunit_assert_collectd_responds_WHEN_sock_connect_fails`, tests a failure to connect to the socket and `testunit_assert_collectd_responds_WHEN_sock_sendcmd_fails` tests a successful connection but fails to send a command after connection.

The ordering may have gotten shuffled around. Would it be clearer to have this one first? Or, what else would help?

monitoring/check_collectd_mlab_test.py
stephen-soltesz: I agree this could be simpler. b0c7825a45913f17779fc56056354d1b4658628f

stephen-soltesz:
Thank you. PTAL?

monitoring/check_collectd_mlab_test.py
mtlynch: Ah, makes sense.

monitoring/check_collectd_mlab_test.py
mtlynch: This looks great, much easier for me to understand.

monitoring/check_collectd_mlab_test.py
mtlynch: I like the distinct names you suggest, but up to you. I don't see it as blocking.

monitoring/check_collectd_mlab_test.py
mtlynch: Oh, I misunderstood. I forgot that "responds" was part of the function name and was reading it as "assert [function] responds when [condition]" and didn't understand the "responds" part.

Withdrawn!

monitoring/check_collectd_mlab_test.py
mtlynch: Ahhhh okay. I was getting lost because there's a few layers of indirection here in mocking out the sock vs. mocking out the return of sock.connect.

One thing I've been trying to do recently when I feel like the test function name isn't clear enough is writing additional explanation in a one-line docstring, because one-line docstrings get printed to the console when the test fails along with the function name. So in this case I'd suggest:

```
  def testunit_assert_collectd_responds_WHEN_sock_sendcmd_fails(
      self, mock_sock_connect):
  """Fail when we create a successful connection, but hit an error on send."""
...

  def testunit_assert_collectd_responds_WHEN_sock_connect_fails(self):
  """Fail if creating a socket connection fails."""
```

monitoring/check_collectd_mlab_test.py
mtlynch: Sorry, I think my earlier comment was unclear. It looks like the current code is producing the wrong result. Right now, callers to `socket.recv` will get back `'1'` then `' '` then `'d'` ...

I think what we want is:

`mock_sock.recv.side_effect = ('1 default reply\n', '')`

monitoring/check_collectd_mlab_test.py
mtlynch: My preference is to use `.return_value` in cases like this when we're always returning a single value because `side_effect` yields unintuitive behavior when we're trying to do something simple. Like here, where I believed this code causes our mocked responses to `recv` to each be 1 character (see my other comment below) even though the code makes it looks like it's returning a string.

Are we deliberately using `.side_effect` because we *want* to simulate behavior where the first call yields the initial answer and subsequent calls fail?

My first choice here would be to just use `return_value` because I don't think it's important to verify `recv` was only called once, but I'm also okay with:

```
mock_sock.recv.side_effect = ('1 default reply\n',)
```

mtlynch:
Splitting up the commits like that makes reviewing the deltas *really* easy. Thanks! Not necessary when it's hard for you to do, but much appreciated when you can.

Finished this round of review. Only blocking issue is I think there's a bug we need to root out with the `.side_effect = list('...')` calls.

monitoring/check_collectd_mlab_test.py
stephen-soltesz: I went for it: 73ec0cca8f72b774b3615425e3272b65b3bd507f

monitoring/check_collectd_mlab_test.py
stephen-soltesz: Ok, I'll keep this in mind. I regularly feel that the test name is not enough.
76421b7ff6b992a8f8d6efb9b50ebfcd8abacf1d

monitoring/check_collectd_mlab_test.py
stephen-soltesz: My understanding of the concern here is that the test code is unintuitive. And, my concern is that we cannot use the unittest.Mock interface in the way you suggest for the current implementation of  `check_collectd_mlab.sock_readline`.

So, I created a `FakeSocketIO` class that is socket-like, supporting `recv` and `send` methods (those used by functions in check_collectd_mlab). Think `StringIO` for sockets. All data to-be-received can be declared initially as a single string, and all sent data can be queried after the test.

I believe this interface is more intuitive because the fake object behaves like an actual socket for the tests code and setup for the FakeSocketIO is straight forward.

How does it look? 301873883cb1227375a48ea6b41b3d57a9c56fa3

So, the arc of this code has been from the FakeCollectd (which while functionally complete was a bit complex) to Mock (which is unintuitive) to FakeSocketIO (which I hope strikes a balance).

monitoring/check_collectd_mlab_test.py
stephen-soltesz: Ok. I think the approach with `FakeSocketIO` is clearer than using a mock now. So, just a clarifying comment on unitest.Mock `.side_effect`.

Originally, I used `.side_effect` because `check_collectd_mlab.sock_readline` calls `recv(1)` which reads one character at a time from the socket. And, the unittest.Mock interface limits our choices to `.return_value` which returns the value "as-is" or `.side_effect` which returns/calls each element in turn. Setting `mock_sock.recv.side_effect = ('1 default reply\n', '')` won't work because `check_collectd_malb.sock_readline` expects one character at a time and on the first call to `recv` the mock will return the entire string `'1 default reply\n'`.

Ideally, I might redesign unittest.Mock to support four attributes, `return_value`, `return_values`, `side_effect`, and `side_effects`, for single or a sequence of plain return values, and single or a sequence of actual "side effects" like exceptions or lambdas. This would prevent the current over-loading on `.side_effect` and cover all common uses cases with less ambiguity.

stephen-soltesz:
Great, I'm glad to hear that splitting up the deltas is helpful.

I think `FakeSocketIO` will address the outstanding concerns about the previous uses of `.side_effect`.

PTAL?

mtlynch:
`FakeSocketIO` greatly improves intelligibility of the tests. I was completely missing the fact that the tests were dependent on 1-byte reads, but this change makes it so that implementation detail doesn't matter (improved readability, reduces flakiness).

LGTM.

PR #12, opened 2015-03-02
Add mlab-view for limited support for collectd-web CGI scripts

stephen-soltesz:
This pull request adds mlab-view, a command line utility for limited support of collectd-web CGI scripts.

This change includes:
* the mlab-view command.
* an update to the collectd-mlab.spec to removes references to httpd-config.py (now combined with mlab-view).
* a minimal httpd.conf, used by mlab-view.
* a collection-mlab.conf to provide human-readable names for custom metrics collected by the mlab plugin.

### DIFF
--- a/viewer/collection-mlab.conf
+++ b/viewer/collection-mlab.conf
@@ -1,0 +1,871 @@
+DataDir "/var/lib/collectd/rrd/"
+GraphWidth 400
+#UnixSockAddr "/var/run/collectd-unixsock"
+<Type apache_bytes>
+  DataSources count
+  DSName "count Bytes/s"
+  RRDTitle "Apache Traffic"
+  RRDVerticalLabel "Bytes/s"
+  RRDFormat "%5.1lf%s"
+  Color count 0000ff
+</Type>
+<Type apache_requests>
+  DataSources count
+  DSName "count Requests/s"
+  RRDTitle "Apache Traffic"
+  RRDVerticalLabel "Requests/s"
+  RRDFormat "%5.2lf"
+  Color count 00d000
+</Type>
+<Type apache_scoreboard>
+  Module GenericStacked
+  DataSources count
+  RRDTitle "Apache scoreboard on {hostname}"
+  RRDVerticalLabel "Slots"
+  RRDFormat "%6.2lf"
+  DSName closing Closing
+  DSName dnslookup DNS lookup
+  DSName finishing Finishing
+  DSName idle_cleanup Idle cleanup
+  DSName keepalive Keep alive
+  DSName logging Logging
+  DSName open Open (empty)
+  DSName reading Reading
+  DSName sending Sending
+  DSName starting Starting
+  DSName waiting Waiting
+  Order open closing dnslookup finishing idle_cleanup keepalive logging open reading sending starting waiting
+  Color closing      000080
+  Color dnslookup    ff0000
+  Color finishing    008080
+  Color idle_cleanup ffff00
+  Color keepalive    0080ff
+  Color logging      a000a0
+  Color open         e0e0e0
+  Color reading      0000ff
+  Color sending      00e000
+  Color starting     ff00ff
+  Color waiting      ffb000
+</Type>
+<Type arc_counts>
+  Module ArcCounts
+  RRDTitle "ARC {type_instance} on {hostname}"
+# RRDOptions ...
+</Type>
+<Type arc_l2_bytes>
+  Module GenericIO
+  DataSources read write
+  DSName  "read Read   "
+  DSName "write Written"
+  RRDTitle "L2ARC traffic"
+  RRDVerticalLabel "Bytes per second"
+# RRDOptions ...
+  RRDFormat "%5.1lf%s"
+</Type>
+<Type arc_l2_size>
+  RRDTitle "L2ARC size on {hostname}"
+  RRDVerticalLabel "Size"
+  RRDFormat "%4.0lf%s"
+  RRDOptions -b 1024
+  DSName "value Current size"
+  Color value  00e000
+</Type>
+<Type arc_size>
+  DataSources "current target minlimit maxlimit"
+  RRDTitle "ARC size on {hostname}"
+  RRDVerticalLabel "Size"
+  RRDFormat "%4.0lf%s"
+  RRDOptions -b 1024
+  DSName  "current Current size"
+  DSName   "target Target size "
+  DSName "maxlimit Max size    "
+  DSName "minlimit Min size    "
+  Color current  00e000
+  Color target   0000ff
+  Color minlimit ff0000
+  Color maxlimit ff00ff
+</Type>
+<Type arc_ratio>
+  DataSources value
+  RRDTitle "{type_instance}ARC ratio on {hostname}"
+  RRDVerticalLabel "Ratio"
+  RRDFormat "%4.1lf"
+  RRDOptions -l 0
+  DSName "value Hit ratio"
+</Type>
+<Type cache_ratio>
+  DataSources value
+  DSName value Percent
+  RRDTitle "Cache hit ratio for {plugin_instance} {type_instance}"
+  RRDVerticalLabel "Percent"
+  RRDFormat "%5.1lf %%"
+</Type>
+<Type cpu>
+  Module GenericStacked
+  DataSources value
+  RRDTitle "CPU {plugin_instance} usage"
+  RRDVerticalLabel "Jiffies"
+  RRDFormat "%5.2lf"
+  RRDOptions -h 512 -w 512 -b 1024 -l 0
+  DSName idle Idle
+  DSName nice Nice
+  DSName user User
+  DSName wait Wait-IO
+  DSName system System
+  DSName softirq SoftIRQ
+  DSName interrupt IRQ
+  DSName steal Steal
+  Order idle nice user wait system softirq interrupt steal
+  Color idle      e8e8e8
+  Color nice      00e000
+  Color user      0000ff
+  Color wait      ffb000
+  Color system    ff0000
+  Color softirq   ff00ff
+  Color interrupt a000a0
+  Color steal     000000
+</Type>
+<Type current>
+  DataSources value
+  DSName value Current
+  RRDTitle "Current ({type_instance})"
+  RRDVerticalLabel "Ampere"
+  RRDFormat "%4.1lfA"
+  Color value ffb000
+</Type>
+<Type df>
+  Module Df
+  DataSources free used
+</Type>
+<Type df_complex>
+  Module GenericStacked
+  DataSources value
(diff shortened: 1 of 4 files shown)


viewer/mlab-view
```diff
@@ -0,0 +1,176 @@
...
+   than the slice IP.
+ * When ssh detects a connection to localhost:8088, the remote end attempts a
+   connection to:
+
+      "utility.mlab.<hostname>:8088"
+
+   To the server, this connection appears to come from the slice IP.
+ * SSH continues to forward all traffic back and forth from localhost:8088
+   through the ssh connection to the remote web server.
+"""
```
mtlynch: Excellent overview, as always.

viewer/mlab-view
```diff
@@ -0,0 +1,176 @@
...
+import time
+
+# Third-party modules.
+import netifaces
+
+
+ACCESS_CONF = '/usr/share/collectd-mlab/access.conf'
+HTTPD_PID = '/etc/httpd/run/httpd.pid'
+
+def log(msg):
```
mtlynch: Need an extra newline before top level function.

viewer/mlab-view
```diff
@@ -0,0 +1,176 @@
...
+  """Sends log message to syslog."""
+  print msg
+  syslog.syslog(syslog.LOG_ERR, 'mlab-view: %s' % msg)
+
+
+def ignore_signal_handler(signal, frame):
```
mtlynch: Does the caller need the parameter to be called `frame`? Otherwise, suggest renaming it to `_` since it is unused.

viewer/mlab-view
```diff
@@ -0,0 +1,176 @@
...
+  """Handler to ignore any signal."""
+  log('Ignoring signal: %s' % signal)
+
+
+def get_address():
+  """Returns local IPv4 & IPv6 addresses as a string separated by spaces."""
+  ifaces = netifaces.interfaces()
+  addrs = []
+  for iface in ifaces:
```
mtlynch: Since we never use `ifaces` again, consider `for iface in netifaces.interfaces():`

viewer/mlab-view
```diff
@@ -0,0 +1,176 @@
...
+
+
+def get_access_config():
+  return """
+<Directory /usr/share/collectd/>
+    AddHandler cgi-script .cgi
+    DirectoryIndex bin/index.cgi
+    Order Deny,Allow
+    Allow from {ips}
+    Deny from all
```
mtlynch: Suggest swapping lines 102 and 103 to be consistent with 101.

viewer/mlab-view
```diff
@@ -0,0 +1,176 @@
...
+
+
+def find_last_parent(pid):
+  """Walks up process tree starting at pid, stopping at sshd or ppid == 1."""
+  try:
+    status_lines = open('/proc/%s/status' % pid, 'r').readlines()
+  except IOError:
+    return -1
+
+  # If the process name contains 'sshd', we'll stop there.
```
mtlynch: This comment seems like it makes more sense on line 125

viewer/mlab-view
```diff
@@ -0,0 +1,176 @@
...
+  name_line = filter(lambda x: x.startswith('Name:'), status_lines)
+  if not name_line:
+    return -1
+
+  ppid_line = filter(lambda x: x.startswith('PPid:'), status_lines)
```
mtlynch: Suggest `name_lines` and `ppid_lines` since these are lists.

viewer/mlab-view
```diff
@@ -0,0 +1,176 @@
...
+    DirectoryIndex bin/index.cgi
+    Order Deny,Allow
+    Allow from {ips}
+    Deny from all
+</Directory>
+""".format(ips=get_address())
+
+
+def find_last_parent(pid):
+  """Walks up process tree starting at pid, stopping at sshd or ppid == 1."""
```
mtlynch: Can we clarify this function a bit? I think "last parent" is kind of ambiguous because it might refer to either to last as in "oldest" or last as in "most recent." Also it would be good to state explicitly what we're returning.

viewer/mlab-view
```diff
@@ -0,0 +1,176 @@
...
+  ppid = int(ppid_line[0].split()[1])
+  if 'sshd' in name_line[0] or ppid == 1:
+    return ppid
+
+  # Continue up the tree.
+  return find_last_parent(ppid)
+
+
+def main():
+  signal.signal(signal.SIGINT, ignore_signal_handler)
```
mtlynch: Can you add a note explaining why we need to ignore these signals?

viewer/mlab-view
```diff
@@ -0,0 +1,176 @@
...
+  except IOError as err:
+    log(err)
+    sys.exit(1)
+
+  if os.path.exists(HTTPD_PID):
+    log('httpd pid file present. Is httpd already running?')
+    sys.exit(1)
+
+  # TODO: what other options are available for serving the cgi files?
+  args = ['/usr/sbin/httpd', '-X', '-f', '/usr/share/collectd-mlab/httpd.conf']
```
mtlynch: I think as long as we're defining path constants at the top, we might as well include these.

viewer/mlab-view
```diff
@@ -0,0 +1,176 @@
...
+import sys
+import syslog
+import time
+
+# Third-party modules.
+import netifaces
+
+
+ACCESS_CONF = '/usr/share/collectd-mlab/access.conf'
+HTTPD_PID = '/etc/httpd/run/httpd.pid'
```
mtlynch: Can we add a note explaining what this is? It's basically a signal that the process is currently running, right?

viewer/collection-mlab.conf
```diff
@@ -0,0 +1,871 @@
...
+  Color rename_table 00d4ff
+  Color replace 00ffff
+  Color revoke 00ffd4
+  Color select 00ffa9
+  Color set_option 00ff7f
+  Color show_create_table 00ff55
+  Color show_databases 00ff2a
+  Color show_fields 00ff00
+  Color show_keys 2aff00
+  Color show_master_status 54ff00
```
mtlynch: Shouldn't this be `53ff00`?

mtlynch: just kidding : )

mtlynch:
Just a few minor notes but LGTM

viewer/mlab-view
stephen-soltesz: Fixed with several cleanups in: 4d941cf3112f5305b7e4fd2fdf2aa64029471c8f

viewer/mlab-view
stephen-soltesz: Renamed unused parameter: 9fc616d81fed5867fd2abb95068138cb13d8c05e

viewer/mlab-view
stephen-soltesz: Fixed with several cleanups in: 4d941cf3112f5305b7e4fd2fdf2aa64029471c8f

viewer/mlab-view
stephen-soltesz: Fixed with several cleanups in: 4d941cf3112f5305b7e4fd2fdf2aa64029471c8f

viewer/mlab-view
stephen-soltesz: Part of the refactor of the process walking function: cbecdfcd61bfcc6b227c2145a591176fc93bcfed

viewer/mlab-view
stephen-soltesz: Good catch. Yeah, I usually use plural names for lists. I'm not sure why I missed these.
Fixed with several cleanups in: 4d941cf3112f5305b7e4fd2fdf2aa64029471c8f

viewer/mlab-view
stephen-soltesz: I agree that this method was unclear. The new version tries to have a better name and includes more notes on what its doing. How does it look? cbecdfcd61bfcc6b227c2145a591176fc93bcfed

viewer/mlab-view
stephen-soltesz: No problem. I'm not entirely clear how signals get routed from a client terminal to the remote process. These ignored signals are attempting to give mlab-view time to shutdown httpd.

The normal mode for mlab-view is when ssh does not allocate a pseudo-tty. In this mode, when a user presses Ctrl-c, the signal is delivered to the ssh command itself, and the corresponding sshd child process on the remote side simply exits. This lets mlab-view clean things up without any signals.

On the other hand, if ssh allocates a pseudo-tty (e.g. by passing "-t") and a user presses Ctrl-c, the signal is sent over the connection. Without ignoring the signal, mlab-view exits and httpd will happily keep running, which we don't want.

What's unclear to me is that ctrl-c (SIGINT) on the client eventually results in mlab-view receiving SIGTERM. So, this signal is ignored for that situation. But, since, I could not find a way to deliver SIGINT and mlab-view worked as expected without that signal ignored in the scenarios I checked, I removed it in: 0478c47103097060d0d40357b89c3aeae3f7bade

Is there a way to improve this?

viewer/mlab-view
stephen-soltesz: No problem. Fixed with several cleanups in: 4d941cf3112f5305b7e4fd2fdf2aa64029471c8f

viewer/mlab-view
stephen-soltesz: That's right. Apache writes that file but does not clean it up. The standard init scripts help with this cleanup, but since we are not using the init system here the cleanup is handled by mlab-view.

viewer/collection-mlab.conf
stephen-soltesz: You actually got me with this.

I was reading your comments in order in the emails sent by github, one comment per email, each one filling up the screen on my phone. I got to this one had no other context and was like "... what on earth!?" Then I read the next email and actually lol'd. :)

stephen-soltesz:
A unified diff for this round: cab330290f58dac92636ca91c8787d7542803e17...0478c47103097060d0d40357b89c3aeae3f7bade

viewer/collection-mlab.conf
mtlynch: Haha

viewer/mlab-view
```diff
@@ -0,0 +1,226 @@
...
+    Deny from all
+    Allow from {ips}
+</Directory>
+""".format(ips=get_address())
+
+
+def is_running(pid):
+  """Checks whether the given pid is running or not."""
+  try:
+    os.kill(pid, 0)
```
mtlynch: Suggest adding a comment making explicit that sending it signal 0 doesn't do anything because for a second I was thinking, "We're checking if it's running by killing it?"

mtlynch:
Updates all look good

viewer/mlab-view
stephen-soltesz: Done. 882c6239dee641d719bb0603f9fb07c7a4660bb6

PR #13, opened 2015-04-12
Increase value range for vs_quota_bytes

stephen-soltesz:
Actual vserver quota values are larger than 32-bit int. This change increases vs_quota_bytes maximum to 64-bit int value.


mtlynch:
LGTM

Issue #14, opened 2015-08-03
collectd crashes


stephen-soltesz:
collectd appears to crash on mlab2.nuq1t.

Two problems:

 * collectd is never restarted.
 * the export script runs unconditionally, which generates reports for all metrics as zero.

Why is collectd crashing?
Add watchdog to restart if collectd is not running? How to handle missing data? How to distinguish between monitoring failure vs system failures?

stephen-soltesz:
collectd was generating a "Bus error" on mlab2.nuq1t and exiting immediately.

 * Set `/etc/default/collectd:DAEMON_COREFILE_LIMIT=unlimited`
 * See trace from gdb below.

Command name abbreviations are allowed if unambiguous.
(gdb) bt
#0  0x00330df9 in __memcpy_ssse3 () from /lib/libc.so.6
#1  0x0011e467 in rrd_write () from /usr/lib/librrd_th.so.4
#2  0x00123c26 in ?? () from /usr/lib/librrd_th.so.4
#3  0x001240dd in ?? () from /usr/lib/librrd_th.so.4
#4  0x001244ab in rrd_update_r () from /usr/lib/librrd_th.so.4
#5  0x00111213 in ?? () from /usr/lib/collectd/rrdtool.so
#6  0x00111aa7 in ?? () from /usr/lib/collectd/rrdtool.so
#7  0x00723b39 in start_thread () from /lib/libpthread.so.0
#8  0x002dec2e in clone () from /lib/libc.so.6

stephen-soltesz:
The node where this was running had a bad disk so I suspected this. But, I also erased the rrd files under `/var/lib/collectd/rrd` wondering if one might be corrupted and collectd has been running since, about 24hrs.

Unfortunately, I erased the files that were corrupted. It would be helpful to have had them available to create a check to notice and prevent this kind of issue from recurring.

Simply restarting collectd will not work if the RRD files are corrupted, because it will crash again.

stephen-soltesz:
A possible hint for detecting corrupted rrds: http://mailman.verplant.org/pipermail/collectd/2009-January/002339.html

The brute force approach might also be to auto-restart collectd, make a note of this, and if there are more than N restarts over 10 minutes, remove all old rrd files so collectd can recreate them.

stephen-soltesz:
To determine if collectd has crashed we might be able to use more specific core patterns.
[...]

PR #17, opened 2015-08-24
Add a collectd watchdog and corrupt RRD cleanup.

stephen-soltesz:
This change addresses issue #14.

Collectd will crash if RRD files become corrupt. This change adds a
watchdog script that will run every five minutes and check if collectd
is running. If it is not, then an integrity check is performed on all
RRD files and collectd is restarted.

This change also adds the watchdog scripts to the collectd-mlab package.


mtlynch:
LGTM

PR #18, opened 2015-08-26
Change host package name.

stephen-soltesz:
The collectd-mlab package consists of two packages -- one for the slice and one for the host.

The host package includes the vsys backend script and the nagios plugin for checking collectd's health.

The previous name only mentioned vsys which is a little confusing.


nkinkade:
LGTM

Issue #20, opened 2015-09-09
/proc/virtual/*/cacct INET* values are inaccurate


stephen-soltesz:
Per-slice network metrics are based on the values in /proc/virtual/<xid>/cacct, but it appears that the values for some slices are incorrect.

The symptom is that aggregate eth0 usage shows all traffic, but slice traffic that should be reported by /proc/virtual/<xid>/cacct is not.

PR #23, opened 2015-10-05
Disable metrics that are known to be unreliable.

stephen-soltesz:
Before enabling log collection, this change disables export of the metrics that were discovered to be unreliable in issue #20.


stephen-soltesz:
LGTM from salarcon@ on code-reviews.

Also bumping collectd-mlab package version number to 1.0.

PR #24, opened 2015-10-26
Add .travis.yml and fix broken tests.

stephen-soltesz:
This change adds a basic travis configuration and fixes unit tests that
were environment-dependent or missed in previous changes.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ language: python
 python:
   - "2.6"
   - "2.7"
-# Install dependencies dependencies
+# Install non-pip dependencies
 before_install:
     - sudo apt-get install -y librrd-dev
 # Install dependencies
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,10 +3,12 @@ python:
   - "2.6"
   - "2.7"
 # Install non-pip dependencies
-addons:
-  apt:
-    packages:
-      - librrd-dev
+before_install:
+  - sudo apt-get install -y librrd-dev
+#addons:
+#  apt:
+#    packages:
+#      - librrd-dev
 # Install dependencies
 install:
     - pip install -r requirements.txt
@@ -18,4 +20,5 @@ script:
     - python -m unittest2 discover -s monitoring/ -p '*_test.py'
     - python -m unittest2 discover -s plugin/ -p '*_test.py'
     - python -m unittest2 discover -s system/vsys/ -p '*_test.py'
-sudo: false
+#sudo: false
+sudo: required
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,8 +3,10 @@ python:
   - "2.6"
   - "2.7"
 # Install non-pip dependencies
-before_install:
-    - sudo apt-get install -y librrd-dev
+addons:
+  apt:
+    sources:
+      - librrd-dev
 # Install dependencies
 install:
     - pip install -r requirements.txt
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,6 @@
 language: python
 python:
+  - "2.6"
   - "2.7"
 # Install dependencies dependencies
 before_install:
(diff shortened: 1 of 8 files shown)


.travis.yml
```diff
@@ -0,0 +1,16 @@
...
+# Install dependencies
+install:
+    - pip install -r requirements.txt
+    - pip install -r test-requirements.txt
+# Run unit tests
+script:
+    - python -m unittest discover -s export/ -p '*_test.py'
+    - python -m unittest discover -s monitoring/ -p '*_test.py'
+    - python -m unittest discover -s plugin/ -p '*_test.py'
+    - python -m unittest discover -s system/vsys/ -p '*_test.py'
```
mtlynch: Add this line at the bottom:

```
sudo: false
```

This declares we don't need sudo and can run on latest-and-greatest Travis infrastructure.

.travis.yml
```diff
@@ -0,0 +1,16 @@
+language: python
+python:
+  - "2.7"
+# Install dependencies dependencies
```
mtlynch: Suggest rephrasing to "non-pip dependencies".

.travis.yml
```diff
@@ -0,0 +1,16 @@
+language: python
+python:
+  - "2.7"
```
mtlynch: I recall during collectd development that it's meant to work on python 2.6 as well, correct? If so, suggest adding a line for "2.6" so that it tests with both runtimes.

export/mlab_export.py
```diff
@@ -302,9 +302,9 @@ def default_output_name(ts_start, ts_end, output_dir):
     str, absolute path of generated output file name.
   """
   filename = '%s-to-%s-metrics.json' % (
-      time.strftime('%Y%m%dT%H:%M:%S', time.localtime(ts_start)),
-      time.strftime('%Y%m%dT%H:%M:%S', time.localtime(ts_end)))
-  date_path = time.strftime('%Y/%m/%d', time.localtime(ts_start))
+      time.strftime('%Y%m%dT%H:%M:%S', time.gmtime(ts_start)),
+      time.strftime('%Y%m%dT%H:%M:%S', time.gmtime(ts_end)))
+  date_path = time.strftime('%Y/%m/%d', time.gmtime(ts_start))
```
mtlynch: Awesome! We're catching bugs already!

mtlynch:
LGTM

.travis.yml
stephen-soltesz: Done. This required a new syntax for the extra package librrd-dev, but it looks like it's passing now.

.travis.yml
stephen-soltesz: Removed this section due to using sudo: false and the new syntax for installing apt packages.

.travis.yml
stephen-soltesz: That's a great idea. Done.

.travis.yml
stephen-soltesz: I'm getting this error. I'm investigating, but I'm not sure how to fix it yet.

The command "python -m unittest2 discover -s export/ -p '*_test.py'" exited with 1.
0.01s$ python -m unittest2 discover -s monitoring/ -p '*_test.py'
/home/travis/virtualenv/python2.6.9/bin/python: unittest2 is a package and cannot be directly executed

mtlynch: Oh looks like the `-m` is a Python 2.7 feature:

http://stackoverflow.com/a/6630931

I'm fine with dropping the 2.6 runtime. I think it's a nice to have but not worth this much hassle.

stephen-soltesz: Nuts. I think it will probably be ok as just 2.7.

stephen-soltesz:
Dropping python 2.6. This might cause some weirdness if we add some 2.7-only features, but they should be caught during early release testing.

Issue #25, opened 2016-04-08
Once lastexport.tstamp is out of sync with current time, it never catches up


stephen-soltesz:
This can result in lost data.

By default, mlab_export calculates the time range to export based on a lastexport.tstamp file. If the machine crashes and mlab_export cannot run, the lastexport timestamp is out of sync with current time.

As a result, mlab_export continues to export one hour at a time from the last known lastexport timestamp, but this will remain out of sync with the current time indefinitely.

mlab_export also creates file names based on the lastexport timestamp, irrespective of current time. So, some data can be written after the collection opportunity has passed. This results in failure to collect in the collection pipeline and missing data. e.g.

    Apr  8 23:10 20160407T16:00:00-to-20160407T17:00:00-metrics.json.gz

mlab_export should instead take the maximum ts_end time based on the current time. This will make the next successful run of mlab_export a "catch up" run. And all subsequent runs will be once an hour as before.

stephen-soltesz:
Should we always use current time to set ts_end?

    if options.ts_end is None:
      alt_end = align_timestamp(align_timestamp(int(time.time()), options.length), options.step)
      def_end = (align_timestamp(options.ts_start, options.step) + options.length)
      options.ts_end = max(alt_end, def_end)
      if options.ts_end == alt_end:
        options.length = options.ts_end - options.ts_start

stephen-soltesz:
This fix will be deployed the week May 16th.

PR #26, opened 2016-04-11
Fix out-of-sync lastexport time stamp

stephen-soltesz:
This PR fixes the last export time stamps growing out of sync by unconditionally setting the ts_end to the nearest hour.

Previously, ts_end was set to the ts_start + length. But, ts_start was set to the previous export time. And, if some delay prevented earlier exports from running (e.g. due to a crash and later reboot), the last export time stamp would never catch up.

This PR fixes issue #25.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/collectd-mlab/26)


mtlynch:
Review status: 0 of 2 files reviewed at latest revision, 4 unresolved discussions.

---

*[export/mlab_export.py, line 261 \[r1\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/26#-KF5ysDRxIiUDYUFM2-t:-KF5ysDRxIiUDYUFM2-u:-1515216974) ([raw file](https://github.com/m-lab/collectd-mlab/blob/ff76da6f53e8c3d3482731f737d2cdd78673b3e7/export/mlab_export.py#L261)):*
Seems like this should be a private method

---

*[export/mlab_export.py, line 266 \[r1\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/26#-KF5yKigJA3y4N_X6Gvl:-KF5yKihDOm6OdvhxUL2:97332755) ([raw file](https://github.com/m-lab/collectd-mlab/blob/ff76da6f53e8c3d3482731f737d2cdd78673b3e7/export/mlab_export.py#L266)):*
Do we need to pass in the entire `options` instance if all we need is `length`?

---

*[export/mlab_export_test.py, line 141 \[r1\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/26#-KF5zORPa4DY8pwePMod:-KF5zORPa4DY8pwePMoe:636208046) ([raw file](https://github.com/m-lab/collectd-mlab/blob/ff76da6f53e8c3d3482731f737d2cdd78673b3e7/export/mlab_export_test.py#L141)):*
I was told it's good practice to always prefer `mock.patch.object` because the interpreter can do better static checking, so I think this would be:

`@mock.patch.object(mlab_export.time, 'time')`

---

*[export/mlab_export_test.py, line 142 \[r1\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/26#-KF5z8VimrOLQCa2uxsJ:-KF5z8VimrOLQCa2uxsK:-193321827) ([raw file](https://github.com/m-lab/collectd-mlab/blob/ff76da6f53e8c3d3482731f737d2cdd78673b3e7/export/mlab_export_test.py#L142)):*
Related to my earlier comment about default_end_time being private, unless we feel like it is a logical part of the class's public interface, I think we should avoid testing it directly and instead test it through `init_args` by mocking out `time.time`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/collectd-mlab/26)*

stephen-soltesz:
Review status: 0 of 2 files reviewed at latest revision, 4 unresolved discussions.

---

*[export/mlab_export.py, line 261 \[r1\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/26#-KF5ysDRxIiUDYUFM2-t:-KF6hoN9ben7KR3v3lDH:-311876744) ([raw file](https://github.com/m-lab/collectd-mlab/blob/ff76da6f53e8c3d3482731f737d2cdd78673b3e7/export/mlab_export.py#L261)):*
I understand. Since this comment applies to almost the entire content of the the collectd-mlab command line tools, I've created an issue to refactor these functions into truly private functions or separate modules and assigned it to myself.

https://github.com/m-lab/collectd-mlab/issues/28

To preserve the current pattern (rather than add one exception), I'll leave this function as-is for now.

---

*[export/mlab_export.py, line 266 \[r1\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/26#-KF5yKigJA3y4N_X6Gvl:-KF6hlzBZCMaKwTfGj3r:1238201172) ([raw file](https://github.com/m-lab/collectd-mlab/blob/ff76da6f53e8c3d3482731f737d2cdd78673b3e7/export/mlab_export.py#L266)):*
Much better suggestion. Done.

---

*[export/mlab_export_test.py, line 141 \[r1\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/26#-KF5zORPa4DY8pwePMod:-KF6hjHc8h3ysDOwiMfw:2061586769) ([raw file](https://github.com/m-lab/collectd-mlab/blob/ff76da6f53e8c3d3482731f737d2cdd78673b3e7/export/mlab_export_test.py#L141)):*
Yes, I agree that the `mock.patch.object` form is preferable. I started by using the form already in the file, but this feels worse than the public/private functions, so I updated the tests I modified.

I've also opened an issue and assigned it to myself to revise all collectd-mlab tests to use the `mock.patch.object()` form, to keep the scope of this PR limited.

https://github.com/m-lab/collectd-mlab/issues/29

---

*[export/mlab_export_test.py, line 142 \[r1\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/26#-KF5z8VimrOLQCa2uxsJ:-KF6kNLuzvdwQ8SZqg14:1123973005) ([raw file](https://github.com/m-lab/collectd-mlab/blob/ff76da6f53e8c3d3482731f737d2cdd78673b3e7/export/mlab_export_test.py#L142)):*
Okay. I think this comment will be addressed by the larger modular refactoring in issue  https://github.com/m-lab/collectd-mlab/issues/28

Ideally, I would like to see interfaces defined well enough that we do not need to mock system libraries for calls that are two levels deep, but instead can pass in a real or fake "time" object for normal operation or testing.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/collectd-mlab/26)*

mtlynch:
:lgtm:

---

Review status: 0 of 2 files reviewed at latest revision, 1 unresolved discussion.

---

*[export/mlab_export.py, line 265 \[r2\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/26#-KF9qG0r-A9TOIcBwumW:-KF9qG0r-A9TOIcBwumX:-1112845892) ([raw file](https://github.com/m-lab/collectd-mlab/blob/51294604193bdd53e39658ff60ed5b9d5c32a2e7/export/mlab_export.py#L265)):*
Is `length` the best name in the context of this function or is it more an artifact of `options`? I can see it making sense, but I also feel like if we were writing this function fresh (as a standalone thing), `length` might not occur to us as a name. Consider `step` to match the parameter in `align_timestamp`.

Also the parameter description reads a little strange to me because it's what the function does with the parameter, not really what the parameter is (though the reader can deduce based on usage). Consider also borrowing the parameter description from `align_timestamp`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/collectd-mlab/26#-:-KF9rcBiJc0QeqFOC1qi:1417088721)*

stephen-soltesz:
Review status: 0 of 2 files reviewed at latest revision, 1 unresolved discussion, all commit checks successful.

---

*[export/mlab_export.py, line 265 \[r2\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/26#-KF9qG0r-A9TOIcBwumW:-KFA47TufhC0I_ceYDQx:1504907949) ([raw file](https://github.com/m-lab/collectd-mlab/blob/51294604193bdd53e39658ff60ed5b9d5c32a2e7/export/mlab_export.py#L265)):*
I think you're right. Updated to `step`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/collectd-mlab/26)*

PR #32, opened 2016-04-14
Add disco python site-package & network module.

stephen-soltesz:
This change introduces a new directory and python package convention for mlab modules. The "mlab.disco" package will be installed in a system site-packages directory so that command line tools may import it easily.

As well, this should establish a pattern for future modules from disco and other tools.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/collectd-mlab/32)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,10 +20,11 @@ script:
     - python -m unittest discover -s monitoring/ -p '*_test.py'
     - python -m unittest discover -s plugin/ -p '*_test.py'
     - python -m unittest discover -s system/vsys/ -p '*_test.py'
+    - python -m unittest discover -s site-packages/ -p '*_test.py'
     # Run auto formatter.
     - yapf --diff --recursive --style google ./ --exclude=./third_party/*
     # Run static analysis for Python bugs/cruft.
     - pyflakes ./
     # Run the docstring lint checker.
-    - PYTHONPATH=$PYTHONPATH:$PWD/third_party/docstringchecker pylint --reports=n plugin/*.py export/*.py monitoring/*.py system/vsys/*.py
+    - PYTHONPATH=$PYTHONPATH:$PWD/third_party/docstringchecker pylint --reports=n plugin/*.py export/*.py monitoring/*.py system/vsys/*.py site-packages/mlab/disco/*.py
 sudo: false
--- a/site-packages/mlab/disco/arp.py
+++ b/site-packages/mlab/disco/arp.py
@@ -6,7 +6,7 @@ class Error(Exception):
     pass
 
 
-class MACNotFound(Exception):
+class MACNotFound(Error):
     """No MAC address is associated with this address."""
     pass
 
@@ -38,7 +38,7 @@ class ArpCache(object):
             fields = line.split()
             self._ip_to_mac[fields[0]] = fields[3]
 
-    def mac_for_ip(self, ip_address):
+    def ip_to_mac(self, ip_address):
         """Returns the MAC address of the given IPv4 address.
 
         Args:
--- a/site-packages/mlab/disco/arp.py
+++ b/site-packages/mlab/disco/arp.py
@@ -1,0 +1,54 @@
+"""Access an ARP cache."""
+
+
+class Error(Exception):
+    """Base error class for arp module."""
+    pass
+
+
+class MACNotFound(Exception):
+    """No MAC address is associated with this address."""
+    pass
+
+# Path to the ARP cache file under the /proc filesystem.
+_ARP_FILENAME = '/proc/net/arp'
+
+
+class ArpCache(object):
+    """"An ARP cache."""
+
+    def __init__(self, arp_lines=None):
+        """Parses an ARP cache.
+
+        Example:
+          $ cat /proc/net/arp
+          IP address       HW type   Flags    HW address          Mask   Device
+          23.228.128.130   0x1       0x2      78:19:f7:86:44:81   *      eth0
+
+        Args:
+          arp_lines: None or list of str, lines from an ARP cache. If None,
+              read lines from the system ARP cache.
+        """
+        if arp_lines is None:
+            with open(_ARP_FILENAME) as arp_file:
+                arp_lines = arp_file.readlines()
+
+        self._ip_to_mac = {}
+        # Skip the header.
+        for line in arp_lines[1:]:
+            fields = line.split()
+            self._ip_to_mac[fields[0]] = fields[3]
+
+    def mac_for_ip(self, ip_address):
+        """Returns the MAC address of the given IPv4 address.
+
+        Args:
+          ip_address: str, IPv4 address to lookup in the ARP cache.
+
+        Raises:
+          MACNotFound: the ip address is not found in the ARP cache.
+        """
+        try:
+            return self._ip_to_mac[ip_address]
+        except KeyError:
+            raise MACNotFound('IP not found in ARP cache: %s' % ip_address)
--- a/site-packages/mlab/disco/arp.py
+++ b/site-packages/mlab/disco/arp.py
@@ -15,7 +15,6 @@ _ARP_FILENAME = '/proc/net/arp'
 
 
 class ArpCache(object):
-    """"An ARP cache."""
 
     def __init__(self, arp_lines=None):
         """Parses an ARP cache.
--- a/site-packages/mlab/disco/arp_test.py
+++ b/site-packages/mlab/disco/arp_test.py
@@ -1,0 +1,35 @@
+"""Tests for arp module."""
+
+import StringIO
+import unittest
+
+# Module under test.
+import arp
+
+FAKE_ARP_CACHE = """
+IP address       HW type     Flags       HW address            Mask     Device
+5.4.3.2          0x1         0x2         00:11:22:33:44:55     *        eth0
+""".strip()
+
+
+class ArpCacheTest(unittest.TestCase):
+    """Tests for the arp.ArpCache class."""
+
+    def test_mac_for_ip(self):
+        arp_cache = arp.ArpCache(StringIO.StringIO(FAKE_ARP_CACHE).readlines())
+        expected_mac = '00:11:22:33:44:55'
+
+        returned_mac = arp_cache.mac_for_ip('5.4.3.2')
+
+        self.assertEqual(expected_mac, returned_mac)
+
+    def test_mac_for_ip_raises_MACNotFound_when_ip_address_not_in_cache(self):
+        arp_cache = arp.ArpCache(StringIO.StringIO(FAKE_ARP_CACHE).readlines())
+
+        with self.assertRaises(arp.MACNotFound):
+            # This address is not found in the FAKE_ARP_CACHE.
+            arp_cache.mac_for_ip('5.4.3.3')
+
+
+if __name__ == '__main__':
+    unittest.main()
--- a/site-packages/mlab/disco/arp_test.py
+++ b/site-packages/mlab/disco/arp_test.py
@@ -14,20 +14,20 @@ IP address       HW type     Flags       HW address            Mask     Device
 class ArpCacheTest(unittest.TestCase):
     """Tests for the arp.ArpCache class."""
 
-    def test_mac_for_ip(self):
+    def test_ip_to_mac(self):
         arp_cache = arp.ArpCache(FAKE_ARP_CACHE.splitlines())
         expected_mac = '00:11:22:33:44:55'
 
-        actual_mac = arp_cache.mac_for_ip('5.4.3.2')
+        actual_mac = arp_cache.ip_to_mac('5.4.3.2')
 
         self.assertEqual(expected_mac, actual_mac)
 
-    def test_mac_for_ip_raises_MACNotFound_when_ip_address_not_in_cache(self):
+    def test_ip_to_mac_raises_MACNotFound_when_ip_address_not_in_cache(self):
         arp_cache = arp.ArpCache(FAKE_ARP_CACHE.splitlines())
 
         with self.assertRaises(arp.MACNotFound):
             # This address is not found in the FAKE_ARP_CACHE.
-            arp_cache.mac_for_ip('5.4.3.3')
+            arp_cache.ip_to_mac('5.4.3.3')
 
 
 if __name__ == '__main__':
--- a/site-packages/mlab/disco/arp_test.py
+++ b/site-packages/mlab/disco/arp_test.py
@@ -1,6 +1,5 @@
 """Tests for arp module."""
 
-import StringIO
 import unittest
 
 # Module under test.
@@ -16,15 +15,15 @@ class ArpCacheTest(unittest.TestCase):
     """Tests for the arp.ArpCache class."""
 
     def test_mac_for_ip(self):
-        arp_cache = arp.ArpCache(StringIO.StringIO(FAKE_ARP_CACHE).readlines())
+        arp_cache = arp.ArpCache(FAKE_ARP_CACHE.splitlines())
         expected_mac = '00:11:22:33:44:55'
 
-        returned_mac = arp_cache.mac_for_ip('5.4.3.2')
+        actual_mac = arp_cache.mac_for_ip('5.4.3.2')
 
-        self.assertEqual(expected_mac, returned_mac)
+        self.assertEqual(expected_mac, actual_mac)
 
     def test_mac_for_ip_raises_MACNotFound_when_ip_address_not_in_cache(self):
-        arp_cache = arp.ArpCache(StringIO.StringIO(FAKE_ARP_CACHE).readlines())
+        arp_cache = arp.ArpCache(FAKE_ARP_CACHE.splitlines())
 
         with self.assertRaises(arp.MACNotFound):
             # This address is not found in the FAKE_ARP_CACHE.
--- a/site-packages/mlab/disco/network.py
+++ b/site-packages/mlab/disco/network.py
@@ -1,5 +1,6 @@
 """Utility methods for accessing local network information."""
 
+import arp
 import netifaces
 
 
@@ -21,14 +22,6 @@ class TooManyInterfaces(Exception):
 class TooFewInterfaces(Exception):
     """Too few physical interfaces were found."""
     pass
-
-
-class UplinkMACNotFound(Exception):
-    """Could not discover a MAC address on switch uplink port."""
-    pass
-
-# Path to the ARP cache file under the /proc filesystem.
-_ARP_FILENAME = '/proc/net/arp'
 
 # Default name for the loopback interface.
 _LOOPBACK = 'lo'
@@ -70,7 +63,7 @@ def get_local_mac():
     return macs.pop()
 
 
-def get_uplink_mac(route_lines=None, arp_lines=None):
+def get_uplink_mac(route_lines=None, arp_cache=None):
     """Discovers a MAC address associated with the switch uplink port.
 
     Multiple MACs may be associated with the switch uplink port. We look
@@ -80,8 +73,8 @@ def get_uplink_mac(route_lines=None, arp_lines=None):
     Args:
       route_lines: None or list of str, lines read from a routing table. If
           None, read lines from the system routing table.
-      arp_lines: None or list of str, lines read from an ARP cache. If None,
-          read lines from the system ARP cache.
+      arp_cache: None or arp.ArpCache, an ARP cache object. If None, use the
+          system ARP cache.
 
     Returns:
       str, a MAC address in six-byte hex form: 00:00:00:00:00:00
@@ -89,23 +82,17 @@ def get_uplink_mac(route_lines=None, arp_lines=None):
     Raises:
       IOError: the routing table or ARP cache files cannot be opened.
       NoDefaultRoute: no default route is found in routing table.
-      UplinkMACNotFound: a MAC address is not found in the ARP cache.
+      arp.MACNotFound: a MAC address is not found in the ARP cache.
     """
     if route_lines is None:
         with open(_ROUTE_FILENAME) as route_file:
             route_lines = route_file.readlines()
 
-    if arp_lines is None:
-        with open(_ARP_FILENAME) as arp_file:
-            arp_lines = arp_file.readlines()
+    if arp_cache is None:
+        arp_cache = arp.ArpCache()
 
     route_ip = _parse_default_route(route_lines)
-    ip_to_mac = _parse_arpcache(arp_lines)
-    try:
-        return ip_to_mac[route_ip]
-    except KeyError:
-        raise UplinkMACNotFound(
-            'Default gateway IP not found in ARP cache: %s' % route_ip)
+    return arp_cache.mac_for_ip(route_ip)
 
 
 def _hexip_to_decip(hex_ip):
@@ -125,7 +112,7 @@ def _hexip_to_decip(hex_ip):
 
 # TODO: Replace with version 0.10+ netifaces.gateways() function when possible.
 def _parse_default_route(route_lines):
-    """Parses a system routing table file, and returns the default gateway IP.
+    """Parses a system routing table, and returns the default gateway IP.
 
     Example:
       $ cat /proc/net/route
@@ -148,26 +135,3 @@ def _parse_default_route(route_lines):
         if fields[1] == _ROUTE_DEFAULT_ADDR:
             return _hexip_to_decip(fields[2])
     raise NoDefaultRoute('No default route found!')
-
-
-def _parse_arpcache(arp_lines):
-    """Parses an ARP cache file, and returns a dict mapping IPs to MACs.
-
-    Example:
-      $ cat /proc/net/arp
-      IP address       HW type   Flags    HW address          Mask   Device
-      23.228.128.130   0x1       0x2      78:19:f7:86:44:81   *      eth0
-
-    Args:
-      arp_lines: list of str, lines from an ARP cache.
-
-    Returns:
-      dict of str, where keys are IP addresses and values are MAC addresses.
-          e.g. {'23.228.128.130': '78:19:f7:86:44:81'}
-    """
-    ip_to_mac = {}
-    # Skip the header.
-    for line in arp_lines[1:]:
-        fields = line.split()
-        ip_to_mac[fields[0]] = fields[3]
-    return ip_to_mac
--- a/site-packages/mlab/disco/network.py
+++ b/site-packages/mlab/disco/network.py
@@ -27,10 +27,16 @@ class UplinkMACNotFound(Exception):
     """Could not discover a MAC address on switch uplink port."""
     pass
 
-
+# Path to the ARP cache file under the /proc filesystem.
 _ARP_FILENAME = '/proc/net/arp'
+
+# Default name for the loopback interface.
 _LOOPBACK = 'lo'
+
+# Path to the routing table file under the /proc filesystem.
 _ROUTE_FILENAME = '/proc/net/route'
+
+# The default IPv4 route address in hexadecimal form.
 _ROUTE_DEFAULT_ADDR = '00000000'
 
 
@@ -41,26 +47,27 @@ def get_local_mac():
       str, local MAC address in six-byte hex form: 00:00:00:00:00:00.
 
     Raises:
-      TooManyInterfaces, if more than one MAC address is discovered.
-      TooFewInterfaces, if no MAC addresses are discovered.
+      TooManyInterfaces: if more than one MAC address is discovered.
+      TooFewInterfaces: if no MAC addresses are discovered.
     """
-    macs = {}
+    macs = set()
 
     for iface in _get_network_interfaces():
         if iface == _LOOPBACK:
             continue
         addrs = _get_interface_addresses(iface)
+        # AF_LINK is a link layer interface, i.e. the physical interface.
         if netifaces.AF_LINK in addrs:
-            macs[addrs[netifaces.AF_LINK][0]['addr']] = 1
+            macs.add(addrs[netifaces.AF_LINK][0]['addr'])
 
     if not macs:
         raise TooFewInterfaces('Could not find any physical interfaces.')
 
     if len(macs) > 1:
         raise TooManyInterfaces(
-            'Can only handle a single physical interface: %s' % macs)
+            'Multiple physical interfaces are not supported, found: %s' % macs)
 
-    return macs.keys()[0]
+    return macs.pop()
 
 
 def get_uplink_mac():
@@ -70,37 +77,40 @@ def get_uplink_mac():
       str, a MAC address in six-byte hex form: 00:00:00:00:00:00
 
     Raises:
-      IOError, if the routing table or arp cache files cannot be opened.
-      NoDefaultRoute, if no default route is found in routing table.
-      UplinkMACNotFound, if a MAC address is not found in the arp cache.
+      IOError: if the routing table or ARP cache files cannot be opened.
+      NoDefaultRoute: if no default route is found in routing table.
+      UplinkMACNotFound: if a MAC address is not found in the ARP cache.
     """
-    routefile = open(_ROUTE_FILENAME, 'r')
-    arpfile = open(_ARP_FILENAME, 'r')
-    return get_uplink_mac_impl(routefile, arpfile)
+    with open(_ROUTE_FILENAME) as route_file:
+        with open(_ARP_FILENAME) as arp_file:
+            return get_uplink_mac_impl(route_file, arp_file)
 
 
-def get_uplink_mac_impl(routefile, arpfile):
+def get_uplink_mac_impl(route_file, arp_file):
     """Discovers a MAC address associated with the switch uplink port.
 
-    Get uplink MAC uses the current routing table and ARP cache.
+    Multiple MACs may be associated with the switch uplink port. We look
+    specifically for the MAC associated with the IPv4 address of the default
+    gateway which is should always be present.
 
     Args:
-      routefile: file, readable file object containing a routing table.
-      arpfile: file, readable file object containing an arp cache.
+      route_file: file, readable file object containing a routing table.
+      arp_file: file, readable file object containing an ARP cache.
 
     Returns:
       str, a MAC address in six-byte hex form: 00:00:00:00:00:00
 
     Raises:
-      NoDefaultRoute, if no default route is found in routing table.
-      UplinkMACNotFound, if a MAC address is not found in the arp cache.
+      NoDefaultRoute: if no default route is found in routing table.
+      UplinkMACNotFound: if a MAC address is not found in the ARP cache.
     """
-    route_ip = _parse_default_route(routefile)
-    arp = _parse_arpcache(arpfile)
-    if route_ip not in arp:
+    route_ip = _parse_default_route(route_file)
+    arp = _parse_arpcache(arp_file)
+    try:
+        return arp[route_ip]
+    except KeyError:
         raise UplinkMACNotFound(
-            'Default gateway IP not found in arp cache: %s' % route_ip)
-    return arp[route_ip]
+            'Default gateway IP not found in ARP cache: %s' % route_ip)
 
 
 def _hexip_to_decip(hex_ip):
@@ -111,7 +121,7 @@ def _hexip_to_decip(hex_ip):
     return '.'.join(dec_fields)
 
 
-def _parse_default_route(routefile):
+def _parse_default_route(route_file):
     """Parses the system routing table, and returns the default gateway IP.
 
     Example:
@@ -120,16 +130,16 @@ def _parse_default_route(routefile):
       eth0    00000000      8180E417  0003    0       0       0       ...
 
     Args:
-      routefile: file, readable file object containing a routing table.
+      route_file: file, readable file object containing a routing table.
 
     Returns:
       str, the decimal form of the destination IP address for the first default
           gateway.
 
     Raises:
-      NoDefaultRoute, if there is no route entry with hex address "00000000".
+      NoDefaultRoute: if there is no route entry with hex address "00000000".
     """
-    cache = routefile.readlines()
+    cache = route_file.readlines()
     # Skip the header.
     for line in cache[1:]:
         fields = line.split()
@@ -138,8 +148,8 @@ def _parse_default_route(routefile):
     raise NoDefaultRoute('No default route found!')
 
 
-def _parse_arpcache(arpfile):
-    """Parses the system arp cache, and returns a dict mapping IPs to MACs.
+def _parse_arpcache(arp_file):
+    """Parses the system ARP cache, and returns a dict mapping IPs to MACs.
 
     Example:
       $ cat /proc/net/arp
@@ -147,12 +157,12 @@ def _parse_arpcache(arpfile):
       23.228.128.130   0x1       0x2      78:19:f7:86:44:81   *      eth0
 
     Args:
-      arpfile: file, readable file object containing an arp cache.
+      arp_file: file, readable file object containing an ARP cache.
 
     Returns:
       dict of str, where keys are IP addresses and values are MAC addresses.
     """
-    cache = arpfile.readlines()
+    cache = arp_file.readlines()
     arp = {}
     # Skip the header.
     for line in cache[1:]:
--- a/site-packages/mlab/disco/network.py
+++ b/site-packages/mlab/disco/network.py
@@ -10,13 +10,18 @@ class Error(Exception):
     pass
 
 
-class TooManyInterfaces(Exception):
+class TooManyInterfaces(Error):
     """Too many physical interfaces were found."""
     pass
 
 
-class TooFewInterfaces(Exception):
+class TooFewInterfaces(Error):
     """Too few physical interfaces were found."""
+    pass
+
+
+class NoDefaultGateway(Error):
+    """Failed to find a default gateway in the routing table."""
     pass
 
 # Default name for the loopback interface.
@@ -69,13 +74,15 @@ def get_uplink_mac(arp_cache=None):
 
     Raises:
       IOError: the routing table or ARP cache files cannot be opened.
-      route.NoDefaultRoute: no default route is found in routing table.
+      NoDefaultGateway: no default route found in routing table.
       arp.MACNotFound: a MAC address is not found in the ARP cache.
     """
-    if arp_cache is None:
+    if not arp_cache:
         arp_cache = arp.ArpCache()
 
     # TODO(soltesz): replace route.gateways with netifaces.gateways when
     # netifaces-0.10 is available in a system package.
     gateways = route.gateways()
-    return arp_cache.mac_for_ip(gateways['default'][route.AF_INET][0])
+    if 'default' not in gateways or route.AF_INET not in gateways['default']:
+        raise NoDefaultGateway('No default route found!')
+    return arp_cache.ip_to_mac(gateways['default'][route.AF_INET][0])
--- a/site-packages/mlab/disco/network.py
+++ b/site-packages/mlab/disco/network.py
@@ -1,0 +1,171 @@
+"""Utility methods for accessing local network information."""
+
+import netifaces
+
+
+class Error(Exception):
+    """Base error class for network module."""
+    pass
+
+
+class NoDefaultRoute(Error):
+    """Failed to find a default route in the routing table."""
+    pass
+
+
+class TooManyInterfaces(Exception):
+    """Too many physical interfaces were found."""
+    pass
+
+
+class TooFewInterfaces(Exception):
+    """Too few physical interfaces were found."""
+    pass
+
+
+class UplinkMACNotFound(Exception):
+    """Could not discover a MAC address on switch uplink port."""
+    pass
+
+
+_ARP_FILENAME = '/proc/net/arp'
+_LOOPBACK = 'lo'
+_ROUTE_FILENAME = '/proc/net/route'
+_ROUTE_DEFAULT_ADDR = '00000000'
+
+
+def get_local_mac():
+    """Discovers the local network interface MAC address.
+
+    Returns:
+      str, local MAC address in six-byte hex form: 00:00:00:00:00:00.
+
+    Raises:
+      TooManyInterfaces, if more than one MAC address is discovered.
+      TooFewInterfaces, if no MAC addresses are discovered.
+    """
+    macs = {}
+
+    for iface in _get_network_interfaces():
+        if iface == _LOOPBACK:
+            continue
+        addrs = _get_interface_addresses(iface)
+        if netifaces.AF_LINK in addrs:
+            macs[addrs[netifaces.AF_LINK][0]['addr']] = 1
+
+    if not macs:
+        raise TooFewInterfaces('Could not find any physical interfaces.')
+
+    if len(macs) > 1:
+        raise TooManyInterfaces(
+            'Can only handle a single physical interface: %s' % macs)
+
+    return macs.keys()[0]
+
+
+def get_uplink_mac():
+    """Wraps get_uplink_mac_impl.
+
+    Returns:
+      str, a MAC address in six-byte hex form: 00:00:00:00:00:00
+
+    Raises:
+      IOError, if the routing table or arp cache files cannot be opened.
+      NoDefaultRoute, if no default route is found in routing table.
+      UplinkMACNotFound, if a MAC address is not found in the arp cache.
+    """
+    routefile = open(_ROUTE_FILENAME, 'r')
+    arpfile = open(_ARP_FILENAME, 'r')
+    return get_uplink_mac_impl(routefile, arpfile)
+
+
+def get_uplink_mac_impl(routefile, arpfile):
+    """Discovers a MAC address associated with the switch uplink port.
+
+    Get uplink MAC uses the current routing table and ARP cache.
+
+    Args:
+      routefile: file, readable file object containing a routing table.
+      arpfile: file, readable file object containing an arp cache.
+
+    Returns:
+      str, a MAC address in six-byte hex form: 00:00:00:00:00:00
+
+    Raises:
+      NoDefaultRoute, if no default route is found in routing table.
+      UplinkMACNotFound, if a MAC address is not found in the arp cache.
+    """
+    route_ip = _parse_default_route(routefile)
+    arp = _parse_arpcache(arpfile)
+    if route_ip not in arp:
+        raise UplinkMACNotFound(
+            'Default gateway IP not found in arp cache: %s' % route_ip)
+    return arp[route_ip]
+
+
+def _hexip_to_decip(hex_ip):
+    """Converts a hexidecimal IP address to decimal form."""
+    fields = [hex_ip[i:i + 2] for i in range(0, len(hex_ip), 2)]
+    dec_fields = [str(int(v, 16)) for v in fields]
+    dec_fields.reverse()
+    return '.'.join(dec_fields)
+
+
+def _parse_default_route(routefile):
+    """Parses the system routing table, and returns the default gateway IP.
+
+    Example:
+      $ cat /proc/net/route
+      Iface   Destination   Gateway   Flags   RefCnt  Use     Metric  ...
+      eth0    00000000      8180E417  0003    0       0       0       ...
+
+    Args:
+      routefile: file, readable file object containing a routing table.
+
+    Returns:
+      str, the decimal form of the destination IP address for the first default
+          gateway.
+
+    Raises:
+      NoDefaultRoute, if there is no route entry with hex address "00000000".
+    """
+    cache = routefile.readlines()
+    # Skip the header.
+    for line in cache[1:]:
+        fields = line.split()
+        if fields[1] == _ROUTE_DEFAULT_ADDR:
+            return _hexip_to_decip(fields[2])
+    raise NoDefaultRoute('No default route found!')
+
+
+def _parse_arpcache(arpfile):
+    """Parses the system arp cache, and returns a dict mapping IPs to MACs.
+
+    Example:
+      $ cat /proc/net/arp
+      IP address       HW type   Flags    HW address          Mask   Device
+      23.228.128.130   0x1       0x2      78:19:f7:86:44:81   *      eth0
+
+    Args:
+      arpfile: file, readable file object containing an arp cache.
+
+    Returns:
+      dict of str, where keys are IP addresses and values are MAC addresses.
+    """
+    cache = arpfile.readlines()
+    arp = {}
+    # Skip the header.
+    for line in cache[1:]:
+        fields = line.split()
+        arp[fields[0]] = fields[3]
+    return arp
+
+
+def _get_network_interfaces():
+    """Wrapper for netifaces.interfaces function."""
+    return netifaces.interfaces()
+
+
+def _get_interface_addresses(iface):
+    """Wrapper for netifaces.ifaddresses function."""
+    return netifaces.ifaddresses(iface)
--- a/site-packages/mlab/disco/network.py
+++ b/site-packages/mlab/disco/network.py
@@ -36,7 +36,7 @@ _LOOPBACK = 'lo'
 # Path to the routing table file under the /proc filesystem.
 _ROUTE_FILENAME = '/proc/net/route'
 
-# The default IPv4 route address in hexadecimal form.
+# The default IPv4 route address as reported by /proc/net/route.
 _ROUTE_DEFAULT_ADDR = '00000000'
 
 
@@ -47,15 +47,15 @@ def get_local_mac():
       str, local MAC address in six-byte hex form: 00:00:00:00:00:00.
 
     Raises:
-      TooManyInterfaces: if more than one MAC address is discovered.
-      TooFewInterfaces: if no MAC addresses are discovered.
+      TooManyInterfaces: more than one MAC address is discovered.
+      TooFewInterfaces: no MAC addresses are discovered.
     """
     macs = set()
 
-    for iface in _get_network_interfaces():
+    for iface in netifaces.interfaces():
         if iface == _LOOPBACK:
             continue
-        addrs = _get_interface_addresses(iface)
+        addrs = netifaces.ifaddresses(iface)
         # AF_LINK is a link layer interface, i.e. the physical interface.
         if netifaces.AF_LINK in addrs:
             macs.add(addrs[netifaces.AF_LINK][0]['addr'])
@@ -70,59 +70,62 @@ def get_local_mac():
     return macs.pop()
 
 
-def get_uplink_mac():
-    """Wraps get_uplink_mac_impl.
-
-    Returns:
-      str, a MAC address in six-byte hex form: 00:00:00:00:00:00
-
-    Raises:
-      IOError: if the routing table or ARP cache files cannot be opened.
-      NoDefaultRoute: if no default route is found in routing table.
-      UplinkMACNotFound: if a MAC address is not found in the ARP cache.
-    """
-    with open(_ROUTE_FILENAME) as route_file:
-        with open(_ARP_FILENAME) as arp_file:
-            return get_uplink_mac_impl(route_file, arp_file)
-
-
-def get_uplink_mac_impl(route_file, arp_file):
+def get_uplink_mac(route_lines=None, arp_lines=None):
     """Discovers a MAC address associated with the switch uplink port.
 
     Multiple MACs may be associated with the switch uplink port. We look
     specifically for the MAC associated with the IPv4 address of the default
-    gateway which is should always be present.
+    gateway which should always be present.
 
     Args:
-      route_file: file, readable file object containing a routing table.
-      arp_file: file, readable file object containing an ARP cache.
+      route_lines: None or list of str, lines read from a routing table. If
+          None, read lines from the system routing table.
+      arp_lines: None or list of str, lines read from an ARP cache. If None,
+          read lines from the system ARP cache.
 
     Returns:
       str, a MAC address in six-byte hex form: 00:00:00:00:00:00
 
     Raises:
-      NoDefaultRoute: if no default route is found in routing table.
-      UplinkMACNotFound: if a MAC address is not found in the ARP cache.
+      IOError: the routing table or ARP cache files cannot be opened.
+      NoDefaultRoute: no default route is found in routing table.
+      UplinkMACNotFound: a MAC address is not found in the ARP cache.
     """
-    route_ip = _parse_default_route(route_file)
-    arp = _parse_arpcache(arp_file)
+    if route_lines is None:
+        with open(_ROUTE_FILENAME) as route_file:
+            route_lines = route_file.readlines()
+
+    if arp_lines is None:
+        with open(_ARP_FILENAME) as arp_file:
+            arp_lines = arp_file.readlines()
+
+    route_ip = _parse_default_route(route_lines)
+    ip_to_mac = _parse_arpcache(arp_lines)
     try:
-        return arp[route_ip]
+        return ip_to_mac[route_ip]
     except KeyError:
         raise UplinkMACNotFound(
             'Default gateway IP not found in ARP cache: %s' % route_ip)
 
 
 def _hexip_to_decip(hex_ip):
-    """Converts a hexidecimal IP address to decimal form."""
+    """Converts a hexadecimal IPv4 address to quad-dotted form.
+
+    Args:
+      hex_ip: str, zero padded, network order, hexadecimal format IPv4 address.
+
+    Returns:
+      str, quad-dotted format IPv4 address.
+    """
     fields = [hex_ip[i:i + 2] for i in range(0, len(hex_ip), 2)]
     dec_fields = [str(int(v, 16)) for v in fields]
     dec_fields.reverse()
     return '.'.join(dec_fields)
 
 
-def _parse_default_route(route_file):
-    """Parses the system routing table, and returns the default gateway IP.
+# TODO: Replace with version 0.10+ netifaces.gateways() function when possible.
+def _parse_default_route(route_lines):
+    """Parses a system routing table file, and returns the default gateway IP.
 
     Example:
       $ cat /proc/net/route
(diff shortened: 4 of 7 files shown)


mtlynch:
Review status: 0 of 3 files reviewed at latest revision, 16 unresolved discussions.

---

*[.travis.yml, line 29 \[r1\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/32#-KFLBxcRPVkbxMqK8uMY:-KFLBxcRPVkbxMqK8uMZ:-1547195334) ([raw file](https://github.com/m-lab/collectd-mlab/blob/de9a933f9cd601415ce677890348de12c39a580b/.travis.yml#L29)):*
(Out of scope because it matches the pattern) Why not have unittest discover and pylint run on everything modulo a blacklist instead of nothing but a whitelist? If we can reduce unittest discover to a single command, then we can trivially add code coverage.

---

*[site-packages/mlab/disco/network.py, line 34 \[r1\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/32#-KFLCNhwP8hNSLfTZ_68:-KFLCNhxnVlP_gqLDv3v:-510754524) ([raw file](https://github.com/m-lab/collectd-mlab/blob/de9a933f9cd601415ce677890348de12c39a580b/site-packages/mlab/disco/network.py#L34)):*
Can we document these?

---

*[site-packages/mlab/disco/network.py, line 53 \[r1\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/32#-KFLFbn3PTUn5xnxGepC:-KFLFbn3PTUn5xnxGepD:-1862023954) ([raw file](https://github.com/m-lab/collectd-mlab/blob/de9a933f9cd601415ce677890348de12c39a580b/site-packages/mlab/disco/network.py#L53)):*
`AF_LINK` means "physical interface"? Suggest stating that as that wasn't immediately obvious to me. Even after looking up [documentation](http://nullege.com/codes/search/netifaces.AF_LINK?fulldoc=1) I'm still not positive.

---

*[site-packages/mlab/disco/network.py, line 54 \[r1\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/32#-KFLF9zKeTkj4NAG5YVy:-KFLF9zKeTkj4NAG5YVz:1806965699) ([raw file](https://github.com/m-lab/collectd-mlab/blob/de9a933f9cd601415ce677890348de12c39a580b/site-packages/mlab/disco/network.py#L54)):*
If all we want is the key portion and we're setting the value to a dummy value, suggest either making `macs` a simple list or a `set`.

---

*[site-packages/mlab/disco/network.py, line 61 \[r1\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/32#-KFLEVDrLUcUCNK1wCmb:-KFLEVDsaR8doLS8rj8e:926084606) ([raw file](https://github.com/m-lab/collectd-mlab/blob/de9a933f9cd601415ce677890348de12c39a580b/site-packages/mlab/disco/network.py#L61)):*
This error message is a bit unclear. Would this indication of a non-sane system or is it a sane system that we just don't support? If it's not supported, suggest rephrasing to `Multiple physical interfaces are not supported, found: %s`

---

*[site-packages/mlab/disco/network.py, line 67 \[r1\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/32#-KFLHUo9FSgesE08ILU4:-KFLHUo9FSgesE08ILU5:-185595160) ([raw file](https://github.com/m-lab/collectd-mlab/blob/de9a933f9cd601415ce677890348de12c39a580b/site-packages/mlab/disco/network.py#L67)):*
Can we add more detail here? Why does `get_uplink_mac_impl` need wrapping?

---

*[site-packages/mlab/disco/network.py, line 73 \[r1\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/32#-KFLIu714u0EYCkFZLGg:-KFLIu714u0EYCkFZLGh:-717084481) ([raw file](https://github.com/m-lab/collectd-mlab/blob/de9a933f9cd601415ce677890348de12c39a580b/site-packages/mlab/disco/network.py#L73)):*
I think these should be like `IOError: The routing table or arp cache files cannot be opened.` to be consistent with the style guide.

https://google.github.io/styleguide/pyguide.html?showone=Comments#Comments

---

*[site-packages/mlab/disco/network.py, line 77 \[r1\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/32#-KFLGHXt7a_olBOMpZ0Q:-KFLGHXt7a_olBOMpZ0R:582703839) ([raw file](https://github.com/m-lab/collectd-mlab/blob/de9a933f9cd601415ce677890348de12c39a580b/site-packages/mlab/disco/network.py#L77)):*
Can we use context handlers here?

---

*[site-packages/mlab/disco/network.py, line 78 \[r1\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/32#-KFLGLDFQgd4i1nyvdZL:-KFLGLDFQgd4i1nyvdZM:-348778723) ([raw file](https://github.com/m-lab/collectd-mlab/blob/de9a933f9cd601415ce677890348de12c39a580b/site-packages/mlab/disco/network.py#L78)):*
(optional) Maybe just a style thing, but I prefer to omit default arguments, especially for really common APIs like `open`

---

*[site-packages/mlab/disco/network.py, line 82 \[r1\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/32#-KFLG3AQ2NdAMBH5xI7P:-KFLG3AQ2NdAMBH5xI7Q:2062760622) ([raw file](https://github.com/m-lab/collectd-mlab/blob/de9a933f9cd601415ce677890348de12c39a580b/site-packages/mlab/disco/network.py#L82)):*
Would we expect a caller to actually call this? It seems strange to expose an `_impl` function to clients.

---

*[site-packages/mlab/disco/network.py, line 83 \[r1\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/32#-KFLIgtruX59AHdcfNOd:-KFLIgtruX59AHdcfNOe:-493508071) ([raw file](https://github.com/m-lab/collectd-mlab/blob/de9a933f9cd601415ce677890348de12c39a580b/site-packages/mlab/disco/network.py#L83)):*
"a MAC" -> "the MAC"?

---

*[site-packages/mlab/disco/network.py, line 85 \[r1\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/32#-KFLHrlV42i0oKX-i-qz:-KFLHrlV42i0oKX-i-r-:1011180917) ([raw file](https://github.com/m-lab/collectd-mlab/blob/de9a933f9cd601415ce677890348de12c39a580b/site-packages/mlab/disco/network.py#L85)):*
Suggest rephrasing to standard description without trying to incorporate the function name. I spent a few seconds trying to parse "Get uplink MAC uses" before I realized "Get uplink MAC" is meant to be a noun phrase because it's the prose version of the function name.

---

*[site-packages/mlab/disco/network.py, line 88 \[r1\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/32#-KFLDFTzgB1thwjq-uri:-KFLDFTzgB1thwjq-urj:1899371367) ([raw file](https://github.com/m-lab/collectd-mlab/blob/de9a933f9cd601415ce677890348de12c39a580b/site-packages/mlab/disco/network.py#L88)):*
(optional) I realize the rest of collectd follows the pattern of specifying types for args and returns, but this feels a little weird to me, like against the spirit of Python. 

I'm not saying that the type should be forbidden, but when we make it mandatory to declare types in the docstring, it feels like we're trying to force a type system onto Python.

---

*[site-packages/mlab/disco/network.py, line 89 \[r1\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/32#-KFLJEqD0D3fl4tdZbZh:-KFLJEqD0D3fl4tdZbZi:-1732548260) ([raw file](https://github.com/m-lab/collectd-mlab/blob/de9a933f9cd601415ce677890348de12c39a580b/site-packages/mlab/disco/network.py#L89)):*
`arp` (not the variable name) should be uppercase

(optional) I'd go `route_file`, `arp_file` to observe the rule that words should be separated by underscore, but I feel like the "infile", "outfile" naming pattern is common enough that it's okay to pretend it's a single word.

---

*[site-packages/mlab/disco/network.py, line 96 \[r1\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/32#-KFLKC9ArnAS8czRjK0p:-KFLKC9ArnAS8czRjK0q:-444208703) ([raw file](https://github.com/m-lab/collectd-mlab/blob/de9a933f9cd601415ce677890348de12c39a580b/site-packages/mlab/disco/network.py#L96)):*
We should uppercase `arp` (I'll stop calling these out individually)

---

*[site-packages/mlab/disco/network.py, line 103 \[r1\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/32#-KFLKXf-avUacm2g3QNc:-KFLKXf-avUacm2g3QNd:1419835724) ([raw file](https://github.com/m-lab/collectd-mlab/blob/de9a933f9cd601415ce677890348de12c39a580b/site-packages/mlab/disco/network.py#L103)):*
Consider restructuring to:

```
try:
  return arp[route_ip]
except KeyError:
  raise UplinkMACNotFound(
    'Default gateway IP not found in arp cache: %s' % route_ip)
```

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/collectd-mlab/32)*

stephen-soltesz:
Review status: 0 of 3 files reviewed at latest revision, 17 unresolved discussions.

---

*[.travis.yml, line 29 \[r1\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/32#-KFLBxcRPVkbxMqK8uMY:-KFLopRC5DwVhjVYI7Hj:-1695206052) ([raw file](https://github.com/m-lab/collectd-mlab/blob/de9a933f9cd601415ce677890348de12c39a580b/.travis.yml#L29)):*
In parallel, I'm making other changes to add coveralls and figure out what it will take to use an auto-discover approach. I want this too, but it seems like much of the current code would need to be organized differently first.

---

*[site-packages/mlab/disco/network.py, line 34 \[r1\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/32#-KFLCNhwP8hNSLfTZ_68:-KFLRE93YIw3iFYNh62Y:2109517944) ([raw file](https://github.com/m-lab/collectd-mlab/blob/de9a933f9cd601415ce677890348de12c39a580b/site-packages/mlab/disco/network.py#L34)):*
Yes. Done.

---

*[site-packages/mlab/disco/network.py, line 53 \[r1\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/32#-KFLFbn3PTUn5xnxGepC:-KFLRexiSaAqeihbY7to:-1423761099) ([raw file](https://github.com/m-lab/collectd-mlab/blob/de9a933f9cd601415ce677890348de12c39a580b/site-packages/mlab/disco/network.py#L53)):*
Good idea. Added a comment.

---

*[site-packages/mlab/disco/network.py, line 54 \[r1\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/32#-KFLF9zKeTkj4NAG5YVy:-KFLoMoKOGA1kZdKjfU5:1976730114) ([raw file](https://github.com/m-lab/collectd-mlab/blob/de9a933f9cd601415ce677890348de12c39a580b/site-packages/mlab/disco/network.py#L54)):*
Oh, yes. A set works nicely.

---

*[site-packages/mlab/disco/network.py, line 61 \[r1\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/32#-KFLEVDrLUcUCNK1wCmb:-KFLS7hbbLrL6TKg657f:1848646273) ([raw file](https://github.com/m-lab/collectd-mlab/blob/de9a933f9cd601415ce677890348de12c39a580b/site-packages/mlab/disco/network.py#L61)):*
Good idea. Changed.

---

*[site-packages/mlab/disco/network.py, line 67 \[r1\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/32#-KFLHUo9FSgesE08ILU4:-KFLoUBAmmR0CZioGcWU:1863526232) ([raw file](https://github.com/m-lab/collectd-mlab/blob/de9a933f9cd601415ce677890348de12c39a580b/site-packages/mlab/disco/network.py#L67)):*
Please see my description on `get_uplink_mac_impl` for what I was going for with testing. Basically, I'm trying to balance "easy to use interface" with "easy to test interface." Happy to explore alternatives.

---

*[site-packages/mlab/disco/network.py, line 73 \[r1\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/32#-KFLIu714u0EYCkFZLGg:-KFLSBBLbPzezgsAQG5R:492752191) ([raw file](https://github.com/m-lab/collectd-mlab/blob/de9a933f9cd601415ce677890348de12c39a580b/site-packages/mlab/disco/network.py#L73)):*
Okay.

:-( Unfortunately, style by example is not as clear or helpful as I would prefer. The pattern in the Args and Returns sections is typically <type><comma><space><description>. In my head, it would be nice for the Raises section to follow the same pattern.

---

*[site-packages/mlab/disco/network.py, line 77 \[r1\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/32#-KFLGHXt7a_olBOMpZ0Q:-KFLd-CoNGgSMVc10GfV:-1274323420) ([raw file](https://github.com/m-lab/collectd-mlab/blob/de9a933f9cd601415ce677890348de12c39a580b/site-packages/mlab/disco/network.py#L77)):*
Ah, yes. Much better.

---

*[site-packages/mlab/disco/network.py, line 78 \[r1\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/32#-KFLGLDFQgd4i1nyvdZL:-KFLdWNqaUWPqCe6G1Ue:-499146009) ([raw file](https://github.com/m-lab/collectd-mlab/blob/de9a933f9cd601415ce677890348de12c39a580b/site-packages/mlab/disco/network.py#L78)):*
Done.

---

*[site-packages/mlab/disco/network.py, line 82 \[r1\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/32#-KFLG3AQ2NdAMBH5xI7P:-KFLddmr1DxJr5VbjyG-:-1840009090) ([raw file](https://github.com/m-lab/collectd-mlab/blob/de9a933f9cd601415ce677890348de12c39a580b/site-packages/mlab/disco/network.py#L82)):*
I separated `get_uplink_mac` and `get_uplink_mac_impl` to improve the tests. My thinking was that `get_uplink_mac` would be obviously correct and we would not need to setup testdata files with opaque contents in the tests. And, instead write really easy to grok tests using `get_uplink_mac_impl`.

What would you recommend?

---

*[site-packages/mlab/disco/network.py, line 83 \[r1\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/32#-KFLIgtruX59AHdcfNOd:-KFLeV07Jx9L6bbgRBAt:-1763040677) ([raw file](https://github.com/m-lab/collectd-mlab/blob/de9a933f9cd601415ce677890348de12c39a580b/site-packages/mlab/disco/network.py#L83)):*
There could be multiple MACs associated with every switch port. This function looks specifically for the IPv4 default gateway's mac, because we know about this one. But, arguably that's an implementation detail? I'm not sure. I updated the docstring to be more explicit. Happy to revise further.

---

*[site-packages/mlab/disco/network.py, line 85 \[r1\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/32#-KFLHrlV42i0oKX-i-qz:-KFLgZT519aErAmXG89K:-891670035) ([raw file](https://github.com/m-lab/collectd-mlab/blob/de9a933f9cd601415ce677890348de12c39a580b/site-packages/mlab/disco/network.py#L85)):*
No problem. Rephrased.

---

*[site-packages/mlab/disco/network.py, line 88 \[r1\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/32#-KFLDFTzgB1thwjq-uri:-KFLgjlH8wWyxT39vzL4:-1330749147) ([raw file](https://github.com/m-lab/collectd-mlab/blob/de9a933f9cd601415ce677890348de12c39a580b/site-packages/mlab/disco/network.py#L88)):*
I'd be happy to talk about this. My preference is to continue specifying the parameter types, mostly to try to disambiguate intention.

---

*[site-packages/mlab/disco/network.py, line 89 \[r1\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/32#-KFLJEqD0D3fl4tdZbZh:-KFLhJJ2ACCl-eqyV3ZG:-926217843) ([raw file](https://github.com/m-lab/collectd-mlab/blob/de9a933f9cd601415ce677890348de12c39a580b/site-packages/mlab/disco/network.py#L89)):*
o_O -- I swear I looked this up and concluded that it could be lower case, but when I look again, clearly someone changed all references to capital letters. Very clever.


Regarding variable name, I think you're right the the underscore is more typically python. In this case, this runs a risk of appearing to place the type specification in the variable name, which I would usually discourage, but "arp" and "route" alone is too vague, I think.

---

*[site-packages/mlab/disco/network.py, line 96 \[r1\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/32#-KFLKC9ArnAS8czRjK0p:-KFLiDpF9Sm7sB9dOnSQ:706291308) ([raw file](https://github.com/m-lab/collectd-mlab/blob/de9a933f9cd601415ce677890348de12c39a580b/site-packages/mlab/disco/network.py#L96)):*
Done throughout.

---

*[site-packages/mlab/disco/network.py, line 103 \[r1\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/32#-KFLKXf-avUacm2g3QNc:-KFLiPZuE5FR1usiNeUc:459906922) ([raw file](https://github.com/m-lab/collectd-mlab/blob/de9a933f9cd601415ce677890348de12c39a580b/site-packages/mlab/disco/network.py#L103)):*
Ah, yes. I like it.

---

*[site-packages/mlab/disco/network.py, line 163 \[r1\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/32#-KFLiY9kUuMoye-cGrCO:-KFLiY9kUuMoye-cGrCP:1581345496) ([raw file](https://github.com/m-lab/collectd-mlab/blob/de9a933f9cd601415ce677890348de12c39a580b/site-packages/mlab/disco/network.py#L163)):*
These are also in the spirit of writing better tests. It may be 50/50 for whether this is preferable to mocking things, but I would like to move toward Fake types rather than deep mocks, if possible. Happy to consider an alternate 'fake' approach here if this isn't working.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/collectd-mlab/32)*

stephen-soltesz:
Review status: 0 of 3 files reviewed at latest revision, 17 unresolved discussions.

---

*[site-packages/mlab/disco/network.py, line 73 \[r1\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/32#-KFLIu714u0EYCkFZLGg:-KFLqMseiFE9RSi8tijL:395552879) ([raw file](https://github.com/m-lab/collectd-mlab/blob/de9a933f9cd601415ce677890348de12c39a580b/site-packages/mlab/disco/network.py#L73)):*
Did not look at the preview. Should read: "The pattern in the Args and Returns sections is typically [type][comma][space][description]."

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/collectd-mlab/32)*

mtlynch:
Reviewed 1 of 3 files at r1, 1 of 2 files at r2, 1 of 1 files at r3.
Review status: all files reviewed at latest revision, 20 unresolved discussions.

---

*[site-packages/mlab/disco/network.py, line 82 \[r1\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/32#-KFLG3AQ2NdAMBH5xI7P:-KFQ8g0zJH3JZ_Q8atGR:1474000615) ([raw file](https://github.com/m-lab/collectd-mlab/blob/de9a933f9cd601415ce677890348de12c39a580b/site-packages/mlab/disco/network.py#L82)):*
My preference would be to make `get_uplink_mac_impl` private and commit the sin of testing a private method. It feels like we're essentially doing that now, but we're just being coy about it by presenting a public method that we know no caller would actually want except for our unit test.

I have an idea for a deeper fix, but it depends on how things go with the other notes, but if we don't need to parse the routes file ourselves, I suggest creating an `ArpCache` class like:

```
class ArpCache(object):

  def __init__(cache_file=None):
    if not cache_file:
      cache_file = open(_ARP_FILENAME) # but w/ context handler, somehow
    self._cache = _parse_cache_file(_ARP_FILENAME)

  def mac_for_ip(ip_address):
    try:
        return self._cache[ip_address]
     except KeyError:
        raise MACNotFound(
            'IP not found in ARP cache: %s' % ip_address)
```

So standard callers will initialize with no param, then test will pass in an explicit mock file.

---

*[site-packages/mlab/disco/network.py, line 83 \[r1\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/32#-KFLIgtruX59AHdcfNOd:-KFPfurD_XfxFizvwA9c:-113784150) ([raw file](https://github.com/m-lab/collectd-mlab/blob/de9a933f9cd601415ce677890348de12c39a580b/site-packages/mlab/disco/network.py#L83)):*
I think the updated docstring clarifies things, thanks!

---

*[site-packages/mlab/disco/network.py, line 163 \[r1\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/32#-KFLiY9kUuMoye-cGrCO:-KFQ7Mner5_0_4TU4rD6:-1487468008) ([raw file](https://github.com/m-lab/collectd-mlab/blob/de9a933f9cd601415ce677890348de12c39a580b/site-packages/mlab/disco/network.py#L163)):*
I think `netifaces` is actually the correct level to mock. The mocks should be the immediate layer below the system under test so that external dependences are explicit. I think mocking as implemented in the current tests is one level too high because now we're chopping out parts of our system under test.

We don't really care if we're calling `_get_network_interfaces` correctly; we care that the implementation ultimately calls the `netifaces` APIs correctly. That way, we can refactor the implementation as much as we want and we won't have to change the test as long as we're still calling `netifaces` correctly.

---

*[site-packages/mlab/disco/network.py, line 40 \[r3\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/32#-KFPgV_pipm1wDaar6_9:-KFPgV_qg9ZOmpCjWpCk:-1028227507) ([raw file](https://github.com/m-lab/collectd-mlab/blob/5dfffd858f81f1584e307d9a5fd566386c968530/site-packages/mlab/disco/network.py#L40)):*
(minor) It's a little confusing to me that this string doesn't match the format we describe in `get_local_mac`'s docstring (`'000000'` vs. `'00:00:00:00:00:00'`) even though we're describing them almost identically as "hexadecimal form." Is there a way for us to be more specific to eliminate the confusion? Like is there a term for the colon separated form or a term for the non-colon form we could use?

---

*[site-packages/mlab/disco/network.py, line 50 \[r3\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/32#-KFPhStTvyUGZRSAJIzi:-KFPhStUC7SF1qqehn0I:-1629517416) ([raw file](https://github.com/m-lab/collectd-mlab/blob/5dfffd858f81f1584e307d9a5fd566386c968530/site-packages/mlab/disco/network.py#L50)):*
(optional) The style guide omits the "if"s. I like this since if we're always going to begin with "if" we might as well just let the "if" be implicit.

---

*[site-packages/mlab/disco/network.py, line 94 \[r3\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/32#-KFPg3nGZXsgJ3OVMIVz:-KFPg3nGZXsgJ3OVMIW-:1416237449) ([raw file](https://github.com/m-lab/collectd-mlab/blob/5dfffd858f81f1584e307d9a5fd566386c968530/site-packages/mlab/disco/network.py#L94)):*
Seems like a stray "is"

---

*[site-packages/mlab/disco/network.py, line 142 \[r3\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/32#-KFQ4n_6EfmSV-AUHB-t:-KFQ4n_6EfmSV-AUHB-u:117896919) ([raw file](https://github.com/m-lab/collectd-mlab/blob/5dfffd858f81f1584e307d9a5fd566386c968530/site-packages/mlab/disco/network.py#L142)):*
Can we lean harder on `netifaces` for this instead of rolling our own?

http://stackoverflow.com/a/24026579/3653712

---

*[site-packages/mlab/disco/network.py, line 152 \[r3\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/32#-KFQ68IItF__Qd-qUcbH:-KFQ68IItF__Qd-qUcbI:1354742743) ([raw file](https://github.com/m-lab/collectd-mlab/blob/5dfffd858f81f1584e307d9a5fd566386c968530/site-packages/mlab/disco/network.py#L152)):*
Suggest rephrasing to "Parses an ARP cache file" since this function parses any cache file, not necessarily the local system's.

---

*[site-packages/mlab/disco/network.py, line 163 \[r3\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/32#-KFQ5xFQvBC_brcSEmEd:-KFQ5xFQvBC_brcSEmEe:1080703297) ([raw file](https://github.com/m-lab/collectd-mlab/blob/5dfffd858f81f1584e307d9a5fd566386c968530/site-packages/mlab/disco/network.py#L163)):*
For complex return types, I think an example is helpful

---

*[site-packages/mlab/disco/network.py, line 166 \[r3\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/32#-KFQAjQ6A5Tlj0XQR2J-:-KFQAjQ6A5Tlj0XQR2J0:-542320261) ([raw file](https://github.com/m-lab/collectd-mlab/blob/5dfffd858f81f1584e307d9a5fd566386c968530/site-packages/mlab/disco/network.py#L166)):*
Suggest `ip_to_arp` (ditto for other places where we use this variable name for the same purpose)

---

*[site-packages/mlab/disco/network_test.py, line 13 \[r3\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/32#-KFPlOCGjXzVHHkFgdNZ:-KFPlOCGjXzVHHkFgdN_:2031154980) ([raw file](https://github.com/m-lab/collectd-mlab/blob/5dfffd858f81f1584e307d9a5fd566386c968530/site-packages/mlab/disco/network_test.py#L13)):*
(optional) I don't think there's actually any benefit to making members of a test module private because a test module shouldn't have any callers, so everything is implicitly private.

---

*[site-packages/mlab/disco/network_test.py, line 32 \[r3\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/32#-KFQBCl7TOq4Alf1xA6z:-KFQBCl8ObFVNl1DD8z2:2100212776) ([raw file](https://github.com/m-lab/collectd-mlab/blob/5dfffd858f81f1584e307d9a5fd566386c968530/site-packages/mlab/disco/network_test.py#L32)):*
This feels a bit dirty. I think it's a symptom of mixing levels of abstraction in the same module.

I feel like if we're testing the logic of "we need to respond in a defined way if the IP from one source doesn't match the IP from another source", it should be at the level of like:

```
test_default_gateway_missing_from_arp_cache_raises_error(self):
  mock_default_gateway = mock.Mock(return_value='1.2.3.4')
  mock_arp_cache = mock.Mock()
  mock_arp_cache.mac_for_ip.side_effect = NotFoundError()
  with self.assertRaises(UplinkMACNotFound):
    find_local_mac(mock_default_gateway, mock_arp_cache))
```

This goes with my suggestion of moving the arp cache logic to a separate module and seeing if we can get rid of the roll-our-own gateway parsing.

---

*[site-packages/mlab/disco/network_test.py, line 39 \[r3\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/32#-KFQ2nVyW7xpGCsTn_hl:-KFQ2nVyW7xpGCsTn_hm:-942008216) ([raw file](https://github.com/m-lab/collectd-mlab/blob/5dfffd858f81f1584e307d9a5fd566386c968530/site-packages/mlab/disco/network_test.py#L39)):*
Consider splitting this into two classes and just have one `TestCase` that patches out those APIs in the `setUp` instead of having to repeatedly add patch decorators.

---

*[site-packages/mlab/disco/network_test.py, line 46 \[r3\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/32#-KFPlp3TM2zIEfBuEAMO:-KFPlp3TM2zIEfBuEAMP:-880151357) ([raw file](https://github.com/m-lab/collectd-mlab/blob/5dfffd858f81f1584e307d9a5fd566386c968530/site-packages/mlab/disco/network_test.py#L46)):*
`_value` is a vague enough suffix that I think it's worth just shortening to `expected` or changing to something more specific like `expected_mac`

---

*[site-packages/mlab/disco/network_test.py, line 52 \[r3\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/32#-KFPmBDw0sfKNzJVq6Vt:-KFPmBDw0sfKNzJVq6Vu:-25737914) ([raw file](https://github.com/m-lab/collectd-mlab/blob/5dfffd858f81f1584e307d9a5fd566386c968530/site-packages/mlab/disco/network_test.py#L52)):*
I think the semantic pairing I'm used to is (expected, actual). This is used in the Python unittest API, though not consistently:

https://docs.python.org/2/library/unittest.html

---

*[site-packages/mlab/disco/network_test.py, line 54 \[r3\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/32#-KFPnjS_qcslgKUr12ZK:-KFPnjS_qcslgKUr12ZL:2091579154) ([raw file](https://github.com/m-lab/collectd-mlab/blob/5dfffd858f81f1584e307d9a5fd566386c968530/site-packages/mlab/disco/network_test.py#L54)):*
(optional) I've recently become won over to the style of introducing redundant strings/values in tests. I think it makes the tests more obvious because there's less to mentally compute in reading the code, so like this would be:

```
mock_network_interfaces.return_value = ['lo', 'eth0']
mock_interface_addresses.return_value = {
  netifaces.AF_LINK: [{'addr': '00:11:22:33:44:55'}]
  }
self.assertEqual('00:11:22:33:44:55', network.get_local_mac())
```

---

*[site-packages/mlab/disco/network_test.py, line 58 \[r3\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/32#-KFQDPHWwu6iqE2le4au:-KFQDPHXUjiIoWJMtdHI:-1151119167) ([raw file](https://github.com/m-lab/collectd-mlab/blob/5dfffd858f81f1584e307d9a5fd566386c968530/site-packages/mlab/disco/network_test.py#L58)):*
The next few test names are a bit confusing. Suggest rephrasing to like:

```
test_get_local_mac_raises_TooFewInterfaces_when_no_network_interfaces_exist
```

---

*[site-packages/mlab/disco/network_test.py, line 73 \[r3\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/32#-KFQ3Vhjqq-15fQJy46X:-KFQ3VhkEDk_4pGlBzcp:777831159) ([raw file](https://github.com/m-lab/collectd-mlab/blob/5dfffd858f81f1584e307d9a5fd566386c968530/site-packages/mlab/disco/network_test.py#L73)):*
Why not `return_value`?

---

*[site-packages/mlab/disco/network_test.py, line 83 \[r3\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/32#-KFQEYrSmvm0MfFlvmk4:-KFQEYrSmvm0MfFlvmk5:-1178342924) ([raw file](https://github.com/m-lab/collectd-mlab/blob/5dfffd858f81f1584e307d9a5fd566386c968530/site-packages/mlab/disco/network_test.py#L83)):*
I renew my request to split into separate modules. The tests for `get_uplink_mac_impl` are pretty hard to follow because it requires jumping around the test file and searching through the entire raw file strings to look for particular MACs/IPs or looking for the *absence* of particular MACs/IPs.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/collectd-mlab/32)*

stephen-soltesz:
Review status: 1 of 3 files reviewed at latest revision, 19 unresolved discussions.

---

*[site-packages/mlab/disco/network.py, line 82 \[r1\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/32#-KFLG3AQ2NdAMBH5xI7P:-KFbMgak-mfoJx5Tu4Rn:1029752446) ([raw file](https://github.com/m-lab/collectd-mlab/blob/de9a933f9cd601415ce677890348de12c39a580b/site-packages/mlab/disco/network.py#L82)):*
Okay, I combined `get_uplink_mac` and `get_uplink_mac_impl` so that they are a single function and `get_uplink_mac` now accepts two optional arguments for the lines read from the route or arp files.

This accomplishes the same goal as the original split without having to test a private function, and at the expense of a more verbose interface to `get_uplink_mac`

How does it look?

---

*[site-packages/mlab/disco/network.py, line 163 \[r1\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/32#-KFLiY9kUuMoye-cGrCO:-KFbFzaFa77a0M5qi43j:-1805496328) ([raw file](https://github.com/m-lab/collectd-mlab/blob/de9a933f9cd601415ce677890348de12c39a580b/site-packages/mlab/disco/network.py#L163)):*
Ok, I've removed these in favor of mocking the netifaces module directly.

---

*[site-packages/mlab/disco/network.py, line 40 \[r3\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/32#-KFPgV_pipm1wDaar6_9:-KFbAIIo20kdKbRsWMZO:1567090678) ([raw file](https://github.com/m-lab/collectd-mlab/blob/5dfffd858f81f1584e307d9a5fd566386c968530/site-packages/mlab/disco/network.py#L40)):*
Okay, I think this constant is actually relative to the `/proc/net/route` file, so I updated the text here to that effect.

And, to be more helpful for other references, I updated the docstring in `_hexip_to_decip` and a few other for consistency.

I found the term "quad-dotted IP" from https://en.wikipedia.org/wiki/IPv4.

---

*[site-packages/mlab/disco/network.py, line 50 \[r3\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/32#-KFPhStTvyUGZRSAJIzi:-KFbBbEepPCtwgrJevCr:1283410147) ([raw file](https://github.com/m-lab/collectd-mlab/blob/5dfffd858f81f1584e307d9a5fd566386c968530/site-packages/mlab/disco/network.py#L50)):*
My argument in favor is that it reads more like a sentence with the 'if': e.g. "get_local_mac raises TooManyInterfaces if more than one MAC address is discovered."

But, Args and Returns entries don't do that, so I've removed the "if".

---

*[site-packages/mlab/disco/network.py, line 94 \[r3\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/32#-KFPg3nGZXsgJ3OVMIVz:-KFRJaKU8IYUBsq4chCe:-335417202) ([raw file](https://github.com/m-lab/collectd-mlab/blob/5dfffd858f81f1584e307d9a5fd566386c968530/site-packages/mlab/disco/network.py#L94)):*
"Please use 'is should' in a sentence: 'Like, such as, is should always be present.'"

Fixed.

---

*[site-packages/mlab/disco/network.py, line 142 \[r3\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/32#-KFQ4n_6EfmSV-AUHB-t:-KFRK6AMj-9X5hm1-9PB:940488937) ([raw file](https://github.com/m-lab/collectd-mlab/blob/5dfffd858f81f1584e307d9a5fd566386c968530/site-packages/mlab/disco/network.py#L142)):*
:-(

The CentOS 6 package for python-netifaces-0.5 only exports the 'ifaddresses' and 'interfaces' functions.

The Ubuntu 14.04 package for python-netifaces-0.8 is the same.

The latest 0.10 version does appear to have the gateways() function.

I've made the argument before that we should prefer system packages, so I've added a TODO to delete this function when we're off of CentOS 6.

---

*[site-packages/mlab/disco/network.py, line 152 \[r3\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/32#-KFQ68IItF__Qd-qUcbH:-KFbDIw-gozRxqnWf6nW:-499146009) ([raw file](https://github.com/m-lab/collectd-mlab/blob/5dfffd858f81f1584e307d9a5fd566386c968530/site-packages/mlab/disco/network.py#L152)):*
Done.

---

*[site-packages/mlab/disco/network.py, line 163 \[r3\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/32#-KFQ5xFQvBC_brcSEmEd:-KFbDWsyUwtHta9kYHsw:1897655028) ([raw file](https://github.com/m-lab/collectd-mlab/blob/5dfffd858f81f1584e307d9a5fd566386c968530/site-packages/mlab/disco/network.py#L163)):*
No problem. Added an example using the "Example" values earlier in the docstring.

---

*[site-packages/mlab/disco/network.py, line 166 \[r3\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/32#-KFQAjQ6A5Tlj0XQR2J-:-KFbDkKy3sTnvvkfcXHb:-1472980052) ([raw file](https://github.com/m-lab/collectd-mlab/blob/5dfffd858f81f1584e307d9a5fd566386c968530/site-packages/mlab/disco/network.py#L166)):*
Updated to `ip_to_mac`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/collectd-mlab/32)*

stephen-soltesz:
Review status: 1 of 5 files reviewed at latest revision, 19 unresolved discussions.

---

*[site-packages/mlab/disco/network_test.py, line 13 \[r3\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/32#-KFPlOCGjXzVHHkFgdNZ:-KFdwFo_-i7ysQUkDpXr:-912862591) ([raw file](https://github.com/m-lab/collectd-mlab/blob/5dfffd858f81f1584e307d9a5fd566386c968530/site-packages/mlab/disco/network_test.py#L13)):*
Okay. Done.

---

*[site-packages/mlab/disco/network_test.py, line 32 \[r3\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/32#-KFQBCl7TOq4Alf1xA6z:-KFe9Tl7BJEXRpgs0Su6:365474665) ([raw file](https://github.com/m-lab/collectd-mlab/blob/5dfffd858f81f1584e307d9a5fd566386c968530/site-packages/mlab/disco/network_test.py#L32)):*
Okay, I added a  new module for "arp.py" that defines the ArpCache class.

And, I'm not comfortable ditching the gateway parsing until netifaces supports this in the system provided package.

---

*[site-packages/mlab/disco/network_test.py, line 39 \[r3\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/32#-KFQ2nVyW7xpGCsTn_hl:-KFdwKc2-tUeOmrj6AW1:-912862591) ([raw file](https://github.com/m-lab/collectd-mlab/blob/5dfffd858f81f1584e307d9a5fd566386c968530/site-packages/mlab/disco/network_test.py#L39)):*
Okay. Done.

---

*[site-packages/mlab/disco/network_test.py, line 46 \[r3\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/32#-KFPlp3TM2zIEfBuEAMO:-KFdwMRPxsC046i3x69Z:-499146009) ([raw file](https://github.com/m-lab/collectd-mlab/blob/5dfffd858f81f1584e307d9a5fd566386c968530/site-packages/mlab/disco/network_test.py#L46)):*
Done.

---

*[site-packages/mlab/disco/network_test.py, line 52 \[r3\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/32#-KFPmBDw0sfKNzJVq6Vt:-KFdwj-DrtMoMjuH0JTp:-419761571) ([raw file](https://github.com/m-lab/collectd-mlab/blob/5dfffd858f81f1584e307d9a5fd566386c968530/site-packages/mlab/disco/network_test.py#L52)):*
Hmm. That is really weird. From some research, it looks like the parameters used to be documented as "actual, expected" but it changed to "first, second", but as you saw there are still some inconsistent usages.

I switched the usage here to "expected, actual" -- is there a reason other than expectation that makes this order easier to remember for you?

---

*[site-packages/mlab/disco/network_test.py, line 54 \[r3\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/32#-KFPnjS_qcslgKUr12ZK:-KFdx7shItd5VSss7sVT:-943228735) ([raw file](https://github.com/m-lab/collectd-mlab/blob/5dfffd858f81f1584e307d9a5fd566386c968530/site-packages/mlab/disco/network_test.py#L54)):*
Sold. Done.

---

*[site-packages/mlab/disco/network_test.py, line 58 \[r3\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/32#-KFQDPHWwu6iqE2le4au:-KFdxqJsiD6gtjF180EZ:-912862591) ([raw file](https://github.com/m-lab/collectd-mlab/blob/5dfffd858f81f1584e307d9a5fd566386c968530/site-packages/mlab/disco/network_test.py#L58)):*
Okay. Done.

---

*[site-packages/mlab/disco/network_test.py, line 73 \[r3\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/32#-KFQ3Vhjqq-15fQJy46X:-KFdy3kE7bp469DwBv19:-1733115131) ([raw file](https://github.com/m-lab/collectd-mlab/blob/5dfffd858f81f1584e307d9a5fd566386c968530/site-packages/mlab/disco/network_test.py#L73)):*
`side_effect` iterates over the elements of the given list for every call to the mocked method. This test needs two return values for when `netifaces.ifaddresses(iface)` is called with `eth0` and `eth1`. `return_value` returns the literal value.

---

*[site-packages/mlab/disco/network_test.py, line 83 \[r3\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/32#-KFQEYrSmvm0MfFlvmk4:-KFeAJUH22AGhpk0aKb6:1177442488) ([raw file](https://github.com/m-lab/collectd-mlab/blob/5dfffd858f81f1584e307d9a5fd566386c968530/site-packages/mlab/disco/network_test.py#L83)):*
Okay, I've tried to split the arp functionality into a separate module. I believe this addresses the current concerns.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/collectd-mlab/32)*

mtlynch:
Reviewed 4 of 4 files at r5.
Review status: all files reviewed at latest revision, 11 unresolved discussions.

---

*[site-packages/mlab/disco/arp.py, line 18 \[r5\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/32#-KFeQ6ClKqZNsehgCT-8:-KFeQ6ClKqZNsehgCT-9:630555225) ([raw file](https://github.com/m-lab/collectd-mlab/blob/7af94a217c3d490f04e35e084633c12d509049c3/site-packages/mlab/disco/arp.py#L18)):*
Docstring doesn't really add much information. Suggest either cutting it or expanding it.

---

*[site-packages/mlab/disco/arp_test.py, line 19 \[r5\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/32#-KFeQeJ1PHUaR1D6Hyh3:-KFeQeJ2Bcm5q5gP_KJO:923855348) ([raw file](https://github.com/m-lab/collectd-mlab/blob/7af94a217c3d490f04e35e084633c12d509049c3/site-packages/mlab/disco/arp_test.py#L19)):*
(minor) `StringIO` seems like overkill if we don't need file-like functionality. Suggest `FAKE_ARP_CACHE.splitlines()`

---

*[site-packages/mlab/disco/network.py, line 82 \[r1\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/32#-KFLG3AQ2NdAMBH5xI7P:-KFeRMtM4MvG_QCKSeNZ:507749641) ([raw file](https://github.com/m-lab/collectd-mlab/blob/de9a933f9cd601415ce677890348de12c39a580b/site-packages/mlab/disco/network.py#L82)):*
I like it!

---

*[site-packages/mlab/disco/network.py, line 142 \[r3\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/32#-KFQ4n_6EfmSV-AUHB-t:-KFeS8JZjBvGg5SCQR3Y:676641074) ([raw file](https://github.com/m-lab/collectd-mlab/blob/5dfffd858f81f1584e307d9a5fd566386c968530/site-packages/mlab/disco/network.py#L142)):*
Durn

---

*[site-packages/mlab/disco/network.py, line 113 \[r5\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/32#-KFeRsDw0C4my5y1uxw7:-KFeRsDw0C4my5y1uxw8:-228281787) ([raw file](https://github.com/m-lab/collectd-mlab/blob/7af94a217c3d490f04e35e084633c12d509049c3/site-packages/mlab/disco/network.py#L113)):*
I think TODOs are always supposed to have a person, even if that person isn't necessarily tasked with the fix:


>TODOs should include the string TODO in all caps, followed by the name, e-mail address, or other identifier of the person who can best provide context about the problem referenced by the TODO, in parentheses. A colon is optional. A comment explaining what there is to do is required. The main purpose is to have a consistent TODO format that can be searched to find the person who can provide more details upon request. A TODO is not a commitment that the person referenced will fix the problem. Thus when you create a TODO, it is almost always your name that is given.

https://google.github.io/styleguide/pyguide.html?showone=TODO_Comments#TODO_Comments

---

*[site-packages/mlab/disco/network_test.py, line 32 \[r3\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/32#-KFQBCl7TOq4Alf1xA6z:-KFeYOr6mbkytU3jPm8v:157616943) ([raw file](https://github.com/m-lab/collectd-mlab/blob/5dfffd858f81f1584e307d9a5fd566386c968530/site-packages/mlab/disco/network_test.py#L32)):*
Yeah, I agree we should use the system provided package. 

Okay, I like the refactoring of the arp.py class. Can we do a similar refactoring for parsing routes? In general, I think parsing is complex and error prone enough that it should always be as tightly encapsulated as possible, so in this case, living in its own separate module.

---

*[site-packages/mlab/disco/network_test.py, line 52 \[r3\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/32#-KFPmBDw0sfKNzJVq6Vt:-KFeTkY34U1vsgxMmaNL:1255245352) ([raw file](https://github.com/m-lab/collectd-mlab/blob/5dfffd858f81f1584e307d9a5fd566386c968530/site-packages/mlab/disco/network_test.py#L52)):*
Oh sorry I was actually talking about naming rather than ordering. I care less about (actual, expected) vs. (expected, actual), though I have a preference for the latter.

Go seems to use the convention of (got, want). .NET always uses (expected, actual). Python seems to use a mix of (expected, actual) and (first, second). I think (first, second) is terrible naming, so I prefer (expected, actual).

---

*[site-packages/mlab/disco/network_test.py, line 73 \[r3\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/32#-KFQ3Vhjqq-15fQJy46X:-KFeTBHVVwGesItoIDuu:1535178634) ([raw file](https://github.com/m-lab/collectd-mlab/blob/5dfffd858f81f1584e307d9a5fd566386c968530/site-packages/mlab/disco/network_test.py#L73)):*
Ahhh gotcha.

---

*[site-packages/mlab/disco/network_test.py, line 35 \[r5\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/32#-KFeTPzTAFmJpBrRKQuL:-KFeTPzTAFmJpBrRKQuM:357679427) ([raw file](https://github.com/m-lab/collectd-mlab/blob/7af94a217c3d490f04e35e084633c12d509049c3/site-packages/mlab/disco/network_test.py#L35)):*
Oh whoa! I didn't realize you could do this. I've been defining it in the `setUp` but this is way cleaner.

---

*[site-packages/mlab/disco/network_test.py, line 75 \[r5\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/32#-KFe_VFSU0cwPL3YoS-d:-KFe_VFSU0cwPL3YoS-e:869327316) ([raw file](https://github.com/m-lab/collectd-mlab/blob/7af94a217c3d490f04e35e084633c12d509049c3/site-packages/mlab/disco/network_test.py#L75)):*
Can we add a success case as well? Also maybe a test where `mac_for_ip` raises `MACNotFound`.

---

*[site-packages/mlab/disco/network_test.py, line 79 \[r5\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/32#-KFeZ3eojyWvr5x9pb7_:-KFeZ3eojyWvr5x9pb7a:-154756328) ([raw file](https://github.com/m-lab/collectd-mlab/blob/7af94a217c3d490f04e35e084633c12d509049c3/site-packages/mlab/disco/network_test.py#L79)):*
I don't follow why we're initializing `arp.ArpCache` since it doesn't look like it will ever get called in this test.

Also, suggest mocking out `arp.ArpCache` in all these tests so that these are explicitly tests of `get_uplink_mac` in isolation. Otherwise the reader has to understand the ARP cache file format to understand this test.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/collectd-mlab/32)*

stephen-soltesz:
Review status: 1 of 7 files reviewed at latest revision, 8 unresolved discussions.

---

*[site-packages/mlab/disco/arp.py, line 18 \[r5\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/32#-KFeQ6ClKqZNsehgCT-8:-KFfO0gDilSDLque5PFO:135510814) ([raw file](https://github.com/m-lab/collectd-mlab/blob/7af94a217c3d490f04e35e084633c12d509049c3/site-packages/mlab/disco/arp.py#L18)):*
Removed. I expected the linter to complain about the missing docstring, but it does not.

---

*[site-packages/mlab/disco/arp_test.py, line 19 \[r5\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/32#-KFeQeJ1PHUaR1D6Hyh3:-KFebZtkjeIY-LXZuexC:-2028301402) ([raw file](https://github.com/m-lab/collectd-mlab/blob/7af94a217c3d490f04e35e084633c12d509049c3/site-packages/mlab/disco/arp_test.py#L19)):*
Ah, yes. That's simpler. A hold over from when we were passing in open files.

---

*[site-packages/mlab/disco/network.py, line 113 \[r5\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/32#-KFeRsDw0C4my5y1uxw7:-KFeax47Mr-FAYRBT6eh:1758359711) ([raw file](https://github.com/m-lab/collectd-mlab/blob/7af94a217c3d490f04e35e084633c12d509049c3/site-packages/mlab/disco/network.py#L113)):*
Fixed.

---

*[site-packages/mlab/disco/network_test.py, line 32 \[r3\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/32#-KFQBCl7TOq4Alf1xA6z:-KFeviIykd4GkjVl8tr7:679318356) ([raw file](https://github.com/m-lab/collectd-mlab/blob/5dfffd858f81f1584e307d9a5fd566386c968530/site-packages/mlab/disco/network_test.py#L32)):*
Hmm. I resisted the idea of pulling such a small function out into a separate module, but then I realized that `_parse_default_route` only exists b/c `netifaces.gateways` is unavailable. So, I've separated this function into a separate module and work-alike implementation of `gateways` that can be a drop-in replaced with `netifaces.gateways` when its available.

---

*[site-packages/mlab/disco/network_test.py, line 52 \[r3\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/32#-KFPmBDw0sfKNzJVq6Vt:-KFec8Ya3PlHJdJB8vbq:11370937) ([raw file](https://github.com/m-lab/collectd-mlab/blob/5dfffd858f81f1584e307d9a5fd566386c968530/site-packages/mlab/disco/network_test.py#L52)):*
Oooh. s/returned/actual/ Done.

---

*[site-packages/mlab/disco/network_test.py, line 35 \[r5\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/32#-KFeTPzTAFmJpBrRKQuL:-KFevToFp6WdNikWdTol:160362572) ([raw file](https://github.com/m-lab/collectd-mlab/blob/7af94a217c3d490f04e35e084633c12d509049c3/site-packages/mlab/disco/network_test.py#L35)):*
:-D

---

*[site-packages/mlab/disco/network_test.py, line 75 \[r5\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/32#-KFe_VFSU0cwPL3YoS-d:-KFeraD2WNDTnz6GXliS:-272423393) ([raw file](https://github.com/m-lab/collectd-mlab/blob/7af94a217c3d490f04e35e084633c12d509049c3/site-packages/mlab/disco/network_test.py#L75)):*
Yes, this was an oversight from the shuffle.

Now with `route.gateways`, the tests here is just a success case.

---

*[site-packages/mlab/disco/network_test.py, line 79 \[r5\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/32#-KFeZ3eojyWvr5x9pb7_:-KFetYzzMnrtwFV_YTQX:750480161) ([raw file](https://github.com/m-lab/collectd-mlab/blob/7af94a217c3d490f04e35e084633c12d509049c3/site-packages/mlab/disco/network_test.py#L79)):*
There's something I'm missing because I feel like this idea keeps coming up in reviews.

From my perspective a mock is less helpful when it is anticipating details of the implementation rather than using fakes that conform to an interface. My thought is that if there is a well defined interface, "fakes" exercise the actual code better than mocks.

For example, the new implementation of `test_get_uplink_mac` uses all mocks to remove the `FAKE_*` route tables and ARP cache files that used to be in the header of `network_test.py`. But, I imagine that this test is very brittle or less helpful for testing the actual interaction between the interfaces in use as part of this test.

This is a genuine question: what am I missing?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/collectd-mlab/32)*

mtlynch:
Reviewed 6 of 6 files at r6.
Review status: all files reviewed at latest revision, 10 unresolved discussions.

---

*[site-packages/mlab/disco/arp.py, line 41 \[r6\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/32#-KFj5YQDpoQtwcqAGFsP:-KFj5YQDpoQtwcqAGFsQ:916576754) ([raw file](https://github.com/m-lab/collectd-mlab/blob/84fc18c7cb223bc626a8db19eac184ee4fcddd4b/site-packages/mlab/disco/arp.py#L41)):*
Sorry, I think I made a bad name suggestion here. `mac_for_ip` is kind of unintuitive because it's "[output] for [input]" whereas the more common pattern is [input] ... [output]". `ip_to_mac`?

---

*[site-packages/mlab/disco/network.py, line 75 \[r6\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/32#-KFiukzXYeF354rNqtGM:-KFiukzXYeF354rNqtGN:1911938711) ([raw file](https://github.com/m-lab/collectd-mlab/blob/84fc18c7cb223bc626a8db19eac184ee4fcddd4b/site-packages/mlab/disco/network.py#L75)):*
This seems like it can just be `if not arp_cache:` since I don't think there's any other falsy possibility for this param.

https://google.github.io/styleguide/pyguide.html?showone=True/False_evaluations#True/False_evaluations

I think this actually applies to all the `if ... is None:` instances in this PR.

---

*[site-packages/mlab/disco/network_test.py, line 79 \[r5\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/32#-KFeZ3eojyWvr5x9pb7_:-KFivS8MjCcP1flkMh1a:-1576746888) ([raw file](https://github.com/m-lab/collectd-mlab/blob/7af94a217c3d490f04e35e084633c12d509049c3/site-packages/mlab/disco/network_test.py#L79)):*
Oh, I see the new implementation as more stable. If we introduce a bug into the ARP cache parsing logic, under the new implementation, it causes only `ArpCacheTest` to fail, whereas the previous implementation would have caused both both `ArpCacheTest` and `NetworkUplinkTest` to fail. I see the previous behavior as undesirable and incorrect because `get_uplink_mac`'s implementation was correct (from the perspective that its job is to be the glue between the default route and the ARP cache), so it shouldn't fail when `ArpCache` breaks.

I also see it as facilitating clearer tests. If `get_uplink_mac`'s job is to glue together the default route and the ARP cache, it's hard for a reader to understand whether it's doing its job correctly if they also have to understand the jobs of its dependencies (parsing the ARP cache and routes files). Under the new implementation, a reader can read this test in isolation and have a pretty good understanding of what `get_uplink_mac` does without understanding any other components of the system. To me, that's a sign of a good unit test.

> From my perspective a mock is less helpful when it is anticipating details of the implementation rather than using fakes that conform to an interface.

Does it, though? To me, `get_uplink_mac` depends on two interfaces: `route.gateways` and `arp_cache.mac_for_ip`. The mocks adhere to those interfaces and don't anticipate any implementation details. The new tests rely on interface contracts to be satisfied, but interface contracts are not implementation details.

---

*[site-packages/mlab/disco/network_test.py, line 29 \[r6\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/32#-KFj4uyWauj202-LW5Y2:-KFj4uyXltxZPKAS67P3:-1716195009) ([raw file](https://github.com/m-lab/collectd-mlab/blob/84fc18c7cb223bc626a8db19eac184ee4fcddd4b/site-packages/mlab/disco/network_test.py#L29)):*
(minor, optional) In instances where we can fit everything on a single line, suggest ditching the additional variables and simplifying to:

```
self.assertEqual('00:11:22:33:44:55', network.get_local_mac())
```

---

*[site-packages/mlab/disco/network_test.py, line 30 \[r6\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/32#-KFj6XiOk3WNOzVOK2nG:-KFj6XiOk3WNOzVOK2nH:1035573659) ([raw file](https://github.com/m-lab/collectd-mlab/blob/84fc18c7cb223bc626a8db19eac184ee4fcddd4b/site-packages/mlab/disco/network_test.py#L30)):*
Would we also want `mock_ifaddresses.assert_called_once_with('eth0')`?

---

*[site-packages/mlab/disco/route.py, line 64 \[r6\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/32#-KFj-VtKa_7OwRCWi-PX:-KFj-VtKa_7OwRCWi-PY:-632311123) ([raw file](https://github.com/m-lab/collectd-mlab/blob/84fc18c7cb223bc626a8db19eac184ee4fcddd4b/site-packages/mlab/disco/route.py#L64)):*
If we want to make this a drop-in replacement for `netifaces.gateways` I think the caller should be responsible for this, otherwise behavior will change when we switch from `route.gateways` to `netifaces.gateways` since `netifaces.gateways` does not consider lack of a default gateway to be an exception.

---

*[site-packages/mlab/disco/route.py, line 67 \[r6\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/32#-KFj-u-opUmY1l68WZkD:-KFj-u-p5uXaL-BEb3ND:-1231004758) ([raw file](https://github.com/m-lab/collectd-mlab/blob/84fc18c7cb223bc626a8db19eac184ee4fcddd4b/site-packages/mlab/disco/route.py#L67)):*
Can we `_` separate instances  of `ip` in the function name? Especially because we're doing it in the param but not the function name. It took me a second to realize `decip` wasn't an abbreviation for `decipher`

---

*[site-packages/mlab/disco/route.py, line 71 \[r6\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/32#-KFj0FSBsdvZoMojIO6u:-KFj0FSCwSFXH4i2B4Rk:908634010) ([raw file](https://github.com/m-lab/collectd-mlab/blob/84fc18c7cb223bc626a8db19eac184ee4fcddd4b/site-packages/mlab/disco/route.py#L71)):*
Can we give an example? (ditto for return address)

---

*[site-packages/mlab/disco/route_test.py, line 12 \[r6\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/32#-KFj1q_bGkLea-M4azFB:-KFj1q_bGkLea-M4azFC:424689304) ([raw file](https://github.com/m-lab/collectd-mlab/blob/84fc18c7cb223bc626a8db19eac184ee4fcddd4b/site-packages/mlab/disco/route_test.py#L12)):*
Suggest throwing in a hex character to make sure we're getting the hex decoding right.

---

*[site-packages/mlab/disco/route_test.py, line 27 \[r6\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/32#-KFj1ZNvCgsmYmslhmV2:-KFj1ZNvCgsmYmslhmV3:1651975628) ([raw file](https://github.com/m-lab/collectd-mlab/blob/84fc18c7cb223bc626a8db19eac184ee4fcddd4b/site-packages/mlab/disco/route_test.py#L27)):*
(minor, optional) I feel like if we're only using it once, might as well define the string locally to the function and save the reader from jumping around the file.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/collectd-mlab/32)*

stephen-soltesz:
Review status: 1 of 7 files reviewed at latest revision, 11 unresolved discussions.

---

*[site-packages/mlab/disco/arp.py, line 9 \[r6\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/32#-KFjMY5RZlpg6DuaGEh2:-KFjMY5RZlpg6DuaGEh3:1812516770) ([raw file](https://github.com/m-lab/collectd-mlab/blob/84fc18c7cb223bc626a8db19eac184ee4fcddd4b/site-packages/mlab/disco/arp.py#L9)):*
(Noticed that a few module exceptions were not inheriting from Error correctly.)

---

*[site-packages/mlab/disco/arp.py, line 41 \[r6\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/32#-KFj5YQDpoQtwcqAGFsP:-KFj7mPHghsNrP5hyNaM:-1062692461) ([raw file](https://github.com/m-lab/collectd-mlab/blob/84fc18c7cb223bc626a8db19eac184ee4fcddd4b/site-packages/mlab/disco/arp.py#L41)):*
No problem. It felt a little funny when I was writing it but didn't think hard enough about it. I agree `ip_to_mac` is better.
Done.

---

*[site-packages/mlab/disco/network.py, line 75 \[r6\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/32#-KFiukzXYeF354rNqtGM:-KFj8GngqtrVUrAkRStb:1790165566) ([raw file](https://github.com/m-lab/collectd-mlab/blob/84fc18c7cb223bc626a8db19eac184ee4fcddd4b/site-packages/mlab/disco/network.py#L75)):*
You're right about this one. `if not arp_cache` is preferable since the arp_cache type shouldn't be something else that can be ambiguously false. (I think it used to be `arp_lines`)

For the instances of `if ... is None` where the parameter is optionally a list, `is None` needs to be distinct from an empty list, and the pattern `if not` does not distinguish them.

---

*[site-packages/mlab/disco/network_test.py, line 79 \[r5\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/32#-KFeZ3eojyWvr5x9pb7_:-KFjLfxVoP0ZCYH2T4AX:1660882094) ([raw file](https://github.com/m-lab/collectd-mlab/blob/7af94a217c3d490f04e35e084633c12d509049c3/site-packages/mlab/disco/network_test.py#L79)):*
Okay. Thank you for the thoughtful reply. I need to keep thinking about this.

---

*[site-packages/mlab/disco/network_test.py, line 29 \[r6\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/32#-KFj4uyWauj202-LW5Y2:-KFjFHlk2grYMaZXvHYP:1652282119) ([raw file](https://github.com/m-lab/collectd-mlab/blob/84fc18c7cb223bc626a8db19eac184ee4fcddd4b/site-packages/mlab/disco/network_test.py#L29)):*
Hmm. I seem to like the "arrange, act, assert" structure when possible to make the "action" (the behavior under test) easier to find. For very short tests this blocking may be less helpful. I've replaced the 'expected' variable with a constant. And, after adding the additional assert, I don't want to "lose" the call to `get_local_mac` inside the assert block, so I left this separate.

---

*[site-packages/mlab/disco/network_test.py, line 30 \[r6\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/32#-KFj6XiOk3WNOzVOK2nG:-KFjIc3kK2lzXK6-gTQP:-1193438574) ([raw file](https://github.com/m-lab/collectd-mlab/blob/84fc18c7cb223bc626a8db19eac184ee4fcddd4b/site-packages/mlab/disco/network_test.py#L30)):*
Yes, I think you're right. Added.

---

*[site-packages/mlab/disco/route.py, line 64 \[r6\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/32#-KFj-VtKa_7OwRCWi-PX:-KFjBW1IkOPzVAJnoi9Q:-117199345) ([raw file](https://github.com/m-lab/collectd-mlab/blob/84fc18c7cb223bc626a8db19eac184ee4fcddd4b/site-packages/mlab/disco/route.py#L64)):*
Yes, good point. Fixed.

---

*[site-packages/mlab/disco/route.py, line 67 \[r6\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/32#-KFj-u-opUmY1l68WZkD:-KFjBiNeJz7mnbUUD6mK:-499146009) ([raw file](https://github.com/m-lab/collectd-mlab/blob/84fc18c7cb223bc626a8db19eac184ee4fcddd4b/site-packages/mlab/disco/route.py#L67)):*
Done.

---

*[site-packages/mlab/disco/route.py, line 71 \[r6\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/32#-KFj0FSBsdvZoMojIO6u:-KFjBx-nqSYUrI7gDcST:-85891810) ([raw file](https://github.com/m-lab/collectd-mlab/blob/84fc18c7cb223bc626a8db19eac184ee4fcddd4b/site-packages/mlab/disco/route.py#L71)):*
Yes we can. Done.

---

*[site-packages/mlab/disco/route_test.py, line 12 \[r6\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/32#-KFj1q_bGkLea-M4azFB:-KFjDLk06McIKlo5-j2s:710635526) ([raw file](https://github.com/m-lab/collectd-mlab/blob/84fc18c7cb223bc626a8db19eac184ee4fcddd4b/site-packages/mlab/disco/route_test.py#L12)):*
Sounds good. Done.

---

*[site-packages/mlab/disco/route_test.py, line 27 \[r6\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/32#-KFj1ZNvCgsmYmslhmV2:-KFjDyjOgRzBFlHRB7xR:-300120753) ([raw file](https://github.com/m-lab/collectd-mlab/blob/84fc18c7cb223bc626a8db19eac184ee4fcddd4b/site-packages/mlab/disco/route_test.py#L27)):*
My concern is the spacing makes this awkward. Either I need to work within the 8-space indent or create a here-doc that ignores the function indent. Since this is optional, I'd prefer to leave the global constants in the header. (Also, this file is pretty small).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/collectd-mlab/32)*

mtlynch:
:lgtm:

---

Reviewed 6 of 6 files at r7.

---

*[site-packages/mlab/disco/route_test.py, line 37 \[r7\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/32#-KFjRok6jboAGgJRhKNg:-KFjRok7L7q0ZMyrt8VZ:-1230948005) ([raw file](https://github.com/m-lab/collectd-mlab/blob/8eba71ada3eb85c2c4214156e62c057b4446c5eb/site-packages/mlab/disco/route_test.py#L37)):*
(optional) Not sure it makes a difference here, but I always assume `assertDictEqual` will do a better job when dealing with dicts.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/collectd-mlab/32#-:-KFjS8ci7kcRYz2ylEJk:1417088721)*

stephen-soltesz:
Thank you!

---

Review status: all files reviewed at latest revision, 2 unresolved discussions, all commit checks successful.

---

*[site-packages/mlab/disco/route_test.py, line 37 \[r7\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/32#-KFjRok6jboAGgJRhKNg:-KFjXa6LYtz65N1TTUZ3:-912862591) ([raw file](https://github.com/m-lab/collectd-mlab/blob/8eba71ada3eb85c2c4214156e62c057b4446c5eb/site-packages/mlab/disco/route_test.py#L37)):*
Okay. Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/collectd-mlab/32#-:-KFjXwXBWQ-1o_7DY0F4:-556193219)*

PR #34, opened 2016-04-20
Add collectd-snmp configuration generation module.

stephen-soltesz:
This PR adds a new module for disco to generate a collectd-snmp configuration file.

### DIFF
--- a/site-packages/mlab/disco/collectd.py
+++ b/site-packages/mlab/disco/collectd.py
@@ -68,17 +68,18 @@ class SnmpConfig(object):
         """Creates a collectd-snmp configuration host block.
 
         Returns:
-          str, a collectd-nsmp <Host></Host> block.
+          str, a collectd-snmp <Host></Host> block.
         """
+        quoted_collections = ['"%s"' % v for v in self._collections]
         # Host name is used to name the output files.
-        data = '<Host "%s">\n' % self._hostname
+        data = '  <Host "%s">\n' % self._hostname
         # Address is used to contact the switch SNMP server.
-        data += '  Address "%s"\n' % self._hostname
-        data += '  Version 2\n'
-        data += '  Community "%s"\n' % self._community
-        data += '  Collect %s\n' % ' '.join(self._collections)
-        data += '  Interval %s\n' % self._interval
-        data += '</Host>'
+        data += '    Address "%s"\n' % self._hostname
+        data += '    Version 2\n'
+        data += '    Community "%s"\n' % self._community
+        data += '    Collect %s\n' % ' '.join(quoted_collections)
+        data += '    Interval %s\n' % self._interval
+        data += '  </Host>'
         return data
 
     def _format_data_block(self, port_name, oid_group_name, oids):
@@ -96,14 +97,14 @@ class SnmpConfig(object):
           tuple of (str, str), the first element is a unique collection name,
               and the second is a collectd-snmp <Data></Data> block.
         """
-        unique_name = _get_unique_name(port_name, oid_group_name)
+        unique_name = self._get_unique_name(port_name, oid_group_name)
 
-        data = '<Data "%s">\n' % unique_name
-        data += '  Type "%s"\n' % _get_collectd_type(oid_group_name)
-        data += '  Table false\n'
-        data += '  Instance "%s"\n' % port_name
-        data += '  Values ' + ' '.join(['"%s"' % v for v in oids]) + '\n'
-        data += '</Data>'
+        data = '  <Data "%s">\n' % unique_name
+        data += '    Type "%s"\n' % self._get_collectd_type(oid_group_name)
+        data += '    Table false\n'
+        data += '    Instance "%s"\n' % port_name
+        data += '    Values ' + ' '.join(['"%s"' % v for v in oids]) + '\n'
+        data += '  </Data>'
         return unique_name, data
 
     def _format_snmp_plugin(self, host_block):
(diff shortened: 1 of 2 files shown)


site-packages/mlab/disco/collectd.py
```diff
@@ -0,0 +1,134 @@
...
+
+
+class SnmpConfig(object):
+    """Configuration for the collectd-snmp plugin."""
+
+    def __init__(self, hostname, community, interval):
+        """Initializes an SnmpConfig object.
+
+        Args:
+          hostname: str, the hostname of the switch SNMP server.
```
nkinkade: nit: I think it would be clearer and more correct to remove "SNMP server"

site-packages/mlab/disco/collectd.py
```diff
@@ -0,0 +1,134 @@
...
+        Args:
+          port_name: str, name of switch port, e.g. 'uplink'.
+          oid_group_name: str, name of the oid to format, e.g. 'discards'.
+          oids: iterable of str, the actual numeric or symbolic OIDs to collect.
+
+        Returns:
+          tuple of (str, str), the first element is a unique collection name,
+              and the second is a collectd-snmp <Data></Data> block.
+        """
+        unique_name = _get_unique_name(port_name, oid_group_name)
```
nkinkade: Should this be **self**._get_unique_name()?

nkinkade: Should this be **self**._get_unique_name()?

site-packages/mlab/disco/collectd.py
```diff
@@ -0,0 +1,134 @@
...
+
+        data = '<Data "%s">\n' % unique_name
+        data += '  Type "%s"\n' % _get_collectd_type(oid_group_name)
```
nkinkade: Similarly, should this be **self**._get_collectd_type()?

site-packages/mlab/disco/collectd.py
```diff
@@ -0,0 +1,134 @@
...
+        Returns:
+          str, a collectd-nsmp <Host></Host> block.
+        """
+        # Host name is used to name the output files.
+        data = '<Host "%s">\n' % self._hostname
+        # Address is used to contact the switch SNMP server.
+        data += '  Address "%s"\n' % self._hostname
+        data += '  Version 2\n'
+        data += '  Community "%s"\n' % self._community
+        data += '  Collect %s\n' % ' '.join(self._collections)
```
nkinkade: Do the space-separated values need to be enclosed in double quotes? I see on line 105 that _Values_ is a space separated list, but each value is enclosed in double quotes.

site-packages/mlab/disco/collectd.py
```diff
@@ -0,0 +1,134 @@
...
+        Returns:
+          str, a collectd-snmp plugin configuration.
+        """
+        return self._format_snmp_plugin(self._format_host_block())
+
+    def _format_host_block(self):
+        """Creates a collectd-snmp configuration host block.
+
+        Returns:
+          str, a collectd-nsmp <Host></Host> block.
```
nkinkade: s/nsmp/snmp/?

site-packages/mlab/disco/collectd.py
```diff
@@ -0,0 +1,134 @@
...
+        Args:
+          host_block: str, a host block as created by format_snmp_host.
+
+        Returns:
+          str, the collectd-snmp configuration.
+        """
+        data = 'LoadPlugin snmp\n'
+        data += '<Plugin snmp>\n'
+        data += '\n'.join(self._data_blocks) + '\n'
+        data += '\n' + host_block + '\n'
```
nkinkade: Could the first '\n' in the above two lines be changed to '\n  ' (newline with two trailing spaces) to achieve indention and better readability within the generated <Plugin/> block?

site-packages/mlab/disco/collectd.py
stephen-soltesz: Hmm, I'd like to leave the "SNMP server" part. I think this is more precise.

site-packages/mlab/disco/collectd.py
stephen-soltesz: Originally, I did not make `_get_unique_name` or `_get_collectd_type` part of the `SnmpConfig` class because they do not depend on the instance, i.e. `self`. But, they're also not especially helpful for generating configurations for non-snmp configs. So, I've included them in the `SnmpConfig` class.

site-packages/mlab/disco/collectd.py
stephen-soltesz: Done.

site-packages/mlab/disco/collectd.py
stephen-soltesz: Good question. Evidently it works without quotes. It was an oversight to omit them however. The examples for the collectd-snmp.conf use quotes. So, I've fixed this here.

https://collectd.org/wiki/index.php/Plugin:SNMP

site-packages/mlab/disco/collectd.py
stephen-soltesz: Done.

site-packages/mlab/disco/collectd.py
stephen-soltesz: Okay. I added extra spaces to the "data" and "host" blocks to get the nested indent. See the sample in collectd_test.py for the net effect.

site-packages/mlab/disco/collectd.py
mtlynch: In code reviews I've gotten, I've been told to not make functions class methods unless they depend on instance variables. I couldn't find anything authoritative to support this, though this [stackoverflow answer](http://stackoverflow.com/questions/11788195/module-function-vs-staticmethod-vs-classmethod-vs-no-decorators-which-idiom-is) supports this.

I think preferring module-level functions is preferable when it doesn't need instance data because it reduces the complexity of the function. In general, we want to give functions access to the lowest amount of input possible, so the function becomes easier to think about if you can instantly see that it has no dependencies outside of its explicit parameters.

site-packages/mlab/disco/collectd.py
nkinkade: Fair enough.

site-packages/mlab/disco/collectd.py
nkinkade: Looks nice now. :) Thanks!

nkinkade:
LGTM.

site-packages/mlab/disco/collectd.py
stephen-soltesz: Sorry, I missed this comment before I merged. I'll create a new PR to change it back.

PR #36, opened 2016-04-21
Add "simple session" module and tests.

stephen-soltesz:
This change adds a simple session class that simplifies the interface to netsnmp.Session. As well, this change adds a fake session class for use in tests in other modules that take a session object.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/collectd-mlab/36)

### DIFF
--- a/site-packages/mlab/disco/simple_session.py
+++ b/site-packages/mlab/disco/simple_session.py
@@ -1,7 +1,17 @@
-"""A simple session interface."""
+"""A simple interface for SNMP sessions."""
 
 import collections
 import netsnmp
+
+
+class Error(Exception):
+    """Base error class for simple_session module."""
+    pass
+
+
+class SNMPError(Error):
+    """A netsnmp error occurred."""
+    pass
 
 
 class SimpleSession(object):
@@ -22,10 +32,17 @@ class SimpleSession(object):
         list is returned.
 
         Returns:
-          list of dict, with each dict containing a 'tag' and 'val' keys.
+          list of dict, with each dict containing 'tag' and 'val' keys.
+
+        Raises:
+          SNMPError: netsnmp session reports any error.
         """
         oids = netsnmp.VarList(netsnmp.Varbind(oid))
         self._session.get(oids)
+        if self._session.ErrorStr:
+            raise SNMPError('%s: %s' % (self._session.ErrorStr,
+                                        self._session.ErrorInd))
+
         return self._varlist_to_list(oids)
 
     def walk(self, oid):
@@ -36,9 +53,15 @@ class SimpleSession(object):
 
         Returns:
           list of dict, with each dict containing a 'tag' and 'val' keys.
+
+        Raises:
+          SNMPError: netsnmp session reports any error.
         """
         oids = netsnmp.VarList(netsnmp.Varbind(oid))
         self._session.walk(oids)
+        if self._session.ErrorStr:
+            raise SNMPError('%s: %s' % (self._session.ErrorStr,
+                                        self._session.ErrorInd))
         return self._varlist_to_list(oids)
 
     def _varlist_to_list(self, oids):
@@ -46,7 +69,7 @@ class SimpleSession(object):
         # The oid.val attribute starts as None and is set after a successful
         # request. If it is still None, then the request failed and it should
         # be excluded from the return value.
-        return [{'tag': oid.tag,
+        return [{'tag': oid.tag + ('.' + oid.iid if oid.iid else ''),
                  'val': oid.val} for oid in oids if oid.val is not None]
 
 
@@ -54,14 +77,14 @@ class FakeSession(object):
     """Implements a fake SimpleSession interface for tests."""
 
     def __init__(self):
-        self._mibs = collections.defaultdict(list)
+        self._mib = collections.defaultdict(list)
 
-    def prepare(self, oid, key, val):
-        """Associates an OID with a key,value for later calls to get or walk."""
-        self._mibs[oid].append({'tag': key, 'val': val})
+    def prepare(self, oid, tag, val):
+        """Associates an OID with a tag,value for later calls to get or walk."""
+        self._mib[oid].append({'tag': tag, 'val': val})
 
-    def get(self, mib):
-        return self._mibs.get(mib, [])
+    def get(self, oid):
+        return self._mib[oid]
 
-    def walk(self, mib):
-        return self._mibs.get(mib, [])
+    def walk(self, oid):
+        return self._mib[oid]
--- a/site-packages/mlab/disco/simple_session.py
+++ b/site-packages/mlab/disco/simple_session.py
@@ -1,6 +1,7 @@
 """A simple interface for SNMP sessions."""
 
 import collections
+# TODO(soltesz): replace netsnmp with easysnmp when available.
 import netsnmp
 
 
@@ -10,8 +11,22 @@ class Error(Exception):
 
 
 class SNMPError(Error):
-    """A netsnmp error occurred."""
+    """A netsnmp error occurred or no results returned."""
     pass
+
+
+# TODO(soltesz): replace SNMPVariable with easysnmp.SNMPVariable when available.
+class SNMPVariable(object):
+    """An SNMP variable representing the result of a get or walk query.
+
+    Attributes:
+        oid: str, the OID of the variable.
+        value: str, the OID value.
+    """
+
+    def __init__(self, oid, value):
+        self.oid = oid
+        self.value = value
 
 
 class SimpleSession(object):
@@ -32,18 +47,14 @@ class SimpleSession(object):
         list is returned.
 
         Returns:
-          list of dict, with each dict containing 'tag' and 'val' keys.
+          list of SNMPVariable, the results of the get request.
 
         Raises:
-          SNMPError: netsnmp session reports any error.
+          SNMPError: netsnmp session reports any error, or no result from get.
         """
         oids = netsnmp.VarList(netsnmp.Varbind(oid))
         self._session.get(oids)
-        if self._session.ErrorStr:
-            raise SNMPError('%s: %s' % (self._session.ErrorStr,
-                                        self._session.ErrorInd))
-
-        return self._varlist_to_list(oids)
+        return self._handle_result(oid, oids)
 
     def walk(self, oid):
         """Reads all values in the MIB tree starting at the given OID.
@@ -52,25 +63,36 @@ class SimpleSession(object):
         list is returned.
 
         Returns:
-          list of dict, with each dict containing a 'tag' and 'val' keys.
+          list of SNMPVariable, the results of the walk request.
 
         Raises:
-          SNMPError: netsnmp session reports any error.
+          SNMPError: netsnmp session reports any error, or no results from walk.
         """
         oids = netsnmp.VarList(netsnmp.Varbind(oid))
         self._session.walk(oids)
+        return self._handle_result(oid, oids)
+
+    def _handle_result(self, oid, oids):
+        """Converts results in oids, or raises exceptions for errors."""
         if self._session.ErrorStr:
-            raise SNMPError('%s: %s' % (self._session.ErrorStr,
-                                        self._session.ErrorInd))
-        return self._varlist_to_list(oids)
+            raise SNMPError('netsnmp error; %s: %s' % (self._session.ErrorStr,
+                                                       self._session.ErrorInd))
+        result = self._varlist_to_list(oids)
+        if not result:
+            raise SNMPError('netsnmp error unknown; OID may be invalid: ' + oid)
+        return result
 
     def _varlist_to_list(self, oids):
-        """Converts a netsnmp.Varlist to a list of dict."""
+        """Converts a netsnmp.Varlist to a list of SNMPVariable."""
         # The oid.val attribute starts as None and is set after a successful
         # request. If it is still None, then the request failed and it should
         # be excluded from the return value.
-        return [{'tag': oid.tag + ('.' + oid.iid if oid.iid else ''),
-                 'val': oid.val} for oid in oids if oid.val is not None]
+        result = []
+        for oid in oids:
+            if oid.val is not None:
+                tag = oid.tag + ('.' + oid.iid if oid.iid else '')
+                result.append(SNMPVariable(tag, oid.val))
+        return result
 
 
 class FakeSession(object):
@@ -79,9 +101,9 @@ class FakeSession(object):
     def __init__(self):
         self._mib = collections.defaultdict(list)
 
-    def prepare(self, oid, tag, val):
+    def prepare(self, oid, tag, value):
         """Associates an OID with a tag,value for later calls to get or walk."""
-        self._mib[oid].append({'tag': tag, 'val': val})
+        self._mib[oid].append(SNMPVariable(tag, value))
 
     def get(self, oid):
         return self._mib[oid]
--- a/site-packages/mlab/disco/simple_session.py
+++ b/site-packages/mlab/disco/simple_session.py
@@ -1,0 +1,67 @@
+"""A simple session interface."""
+
+import collections
+import netsnmp
+
+
+class SimpleSession(object):
+    """A simpler session interface for netsnmp.Session."""
+
+    def __init__(self, session):
+        """Creates a SimpleSession object.
+
+        Args:
+          session: netsnmp.Session, the SNMP session to use for queries.
+        """
+        self._session = session
+
+    def get(self, oid):
+        """Reads the value at the given OID.
+
+        If an error occurs such that the get could not be completed, an empty
+        list is returned.
+
+        Returns:
+          list of dict, with each dict containing a 'tag' and 'val' keys.
+        """
+        oids = netsnmp.VarList(netsnmp.Varbind(oid))
+        self._session.get(oids)
+        return self._varlist_to_list(oids)
+
+    def walk(self, oid):
+        """Reads all values in the MIB tree starting at the given OID.
+
+        If an error occurs such that the walk could not be completed, an empty
+        list is returned.
+
+        Returns:
+          list of dict, with each dict containing a 'tag' and 'val' keys.
+        """
+        oids = netsnmp.VarList(netsnmp.Varbind(oid))
+        self._session.walk(oids)
+        return self._varlist_to_list(oids)
+
+    def _varlist_to_list(self, oids):
+        """Converts a netsnmp.Varlist to a list of dict."""
+        # The oid.val attribute starts as None and is set after a successful
+        # request. If it is still None, then the request failed and it should
+        # be excluded from the return value.
+        return [{'tag': oid.tag,
+                 'val': oid.val} for oid in oids if oid.val is not None]
+
+
+class FakeSession(object):
+    """Implements a fake SimpleSession interface for tests."""
+
+    def __init__(self):
+        self._mibs = collections.defaultdict(list)
+
+    def prepare(self, oid, key, val):
+        """Associates an OID with a key,value for later calls to get or walk."""
+        self._mibs[oid].append({'tag': key, 'val': val})
+
+    def get(self, mib):
+        return self._mibs.get(mib, [])
+
+    def walk(self, mib):
+        return self._mibs.get(mib, [])
--- a/site-packages/mlab/disco/simple_session_test.py
+++ b/site-packages/mlab/disco/simple_session_test.py
@@ -10,21 +10,17 @@ import mock
 import simple_session
 
 
-def make_varbind(tag, val):
-    var = netsnmp.Varbind(tag)
-    var.val = val
-    return var
-
-
 @mock.patch.object(netsnmp, 'Varbind')
 @mock.patch.object(netsnmp, 'VarList')
 class SimpleSessionTest(unittest.TestCase):
 
     def setUp(self):
-        self.var = make_varbind('sysDescr.0', 'Fake system description.')
+        self.var = netsnmp.Varbind('sysDescr.0', val='Fake system description.')
         self.var_list = [
-            make_varbind('BRIDGE-MIB::dot1dTpFdbPort.0.12.34.56.78.150', '2'),
-            make_varbind('BRIDGE-MIB::dot1dTpFdbPort.0.12.34.56.78.160', '1'),
+            netsnmp.Varbind('BRIDGE-MIB::dot1dTpFdbPort.0.12.34.56.78.150',
+                            val='2'),
+            netsnmp.Varbind('BRIDGE-MIB::dot1dTpFdbPort.0.12.34.56.78.160',
+                            val='1'),
         ]
         self.mock_session = mock.Mock(spec=netsnmp.Session)
         self.mock_session.ErrorStr = None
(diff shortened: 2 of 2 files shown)


mtlynch:
Review status: 0 of 2 files reviewed at latest revision, 9 unresolved discussions.

---

*[site-packages/mlab/disco/simple_session.py, line 1 \[r1\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/36#-KFwyC2Qk5OIFhc4eiSX:-KFwyC2Qk5OIFhc4eiSY:-2055987815) ([raw file](https://github.com/m-lab/collectd-mlab/blob/c0626d33ad3e68860c00a697e638b2c254b401d3/site-packages/mlab/disco/simple_session.py#L1)):*
This feels a bit too brief and doesn't give much information beyond what the reader would guess based on the filename. "A simple interface for SNMP sessions"?

---

*[site-packages/mlab/disco/simple_session.py, line 22 \[r1\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/36#-KFx0-CUEwVICwj90NUU:-KFx0-CV4WEDQmtJFNq9:-1628222271) ([raw file](https://github.com/m-lab/collectd-mlab/blob/c0626d33ad3e68860c00a697e638b2c254b401d3/site-packages/mlab/disco/simple_session.py#L22)):*
It seems strange to give the caller no way to get any information about the error. Does it make sense for us to detect when an error occurs and then raise an exception based on the `ErrorStr` value?

---

*[site-packages/mlab/disco/simple_session.py, line 25 \[r1\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/36#-KFwyWTujXkeUdV1NUYp:-KFwyWTvZ5zXLVjwDAgq:451587738) ([raw file](https://github.com/m-lab/collectd-mlab/blob/c0626d33ad3e68860c00a697e638b2c254b401d3/site-packages/mlab/disco/simple_session.py#L25)):*
I feel like passing around dicts instead of defining classes is generally more confusing. Can we define an `SnmpVariable` class (or whatever the appropriate name for this structure would be)?

If not, extraneous "a" after "containing"

---

*[site-packages/mlab/disco/simple_session.py, line 57 \[r1\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/36#-KFx384bLQXI1WaSVbHr:-KFx384bLQXI1WaSVbHs:-1684084845) ([raw file](https://github.com/m-lab/collectd-mlab/blob/c0626d33ad3e68860c00a697e638b2c254b401d3/site-packages/mlab/disco/simple_session.py#L57)):*
`_mibs` is an unexpected name because we're not using it anywhere else. `_varbinds`?

---

*[site-packages/mlab/disco/simple_session.py, line 59 \[r1\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/36#-KFx0VeEZIUR2c1w822C:-KFx0VeEZIUR2c1w822D:1197006229) ([raw file](https://github.com/m-lab/collectd-mlab/blob/c0626d33ad3e68860c00a697e638b2c254b401d3/site-packages/mlab/disco/simple_session.py#L59)):*
I think we should be consistent and use either (key, val) or (tag, val) throughout. Probably tag, val if that's what netsnmp uses.

---

*[site-packages/mlab/disco/simple_session.py, line 64 \[r1\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/36#-KFx2rjhENF3d23E4bKK:-KFx2rjiaNmKrIWvQ83Y:1760317729) ([raw file](https://github.com/m-lab/collectd-mlab/blob/c0626d33ad3e68860c00a697e638b2c254b401d3/site-packages/mlab/disco/simple_session.py#L64)):*
It seems like we're treating it as a normal `dict` instead of a `defaultdict`. Won't `self._mibs[mib]` do what we want?

---

*[site-packages/mlab/disco/simple_session_test.py, line 14 \[r1\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/36#-KFx28cH3ZqGoT3CeRZO:-KFx28cH3ZqGoT3CeRZP:2140501849) ([raw file](https://github.com/m-lab/collectd-mlab/blob/c0626d33ad3e68860c00a697e638b2c254b401d3/site-packages/mlab/disco/simple_session_test.py#L14)):*
Why not use a real one?

---

*[site-packages/mlab/disco/simple_session_test.py, line 39 \[r1\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/36#-KFx1Zbae8E-5nQi8HAy:-KFx1ZbbkIiX5Fh6KYi6:701160158) ([raw file](https://github.com/m-lab/collectd-mlab/blob/c0626d33ad3e68860c00a697e638b2c254b401d3/site-packages/mlab/disco/simple_session_test.py#L39)):*
Can we add a test for the error case?

---

*[site-packages/mlab/disco/simple_session_test.py, line 52 \[r1\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/36#-KFx0zpNc0lSMc0jekgb:-KFx0zpNc0lSMc0jekgc:391736232) ([raw file](https://github.com/m-lab/collectd-mlab/blob/c0626d33ad3e68860c00a697e638b2c254b401d3/site-packages/mlab/disco/simple_session_test.py#L52)):*
I think we could use more exhaustive tests for `walk` like if it results in a list > 1 and if it results in a list with errors.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/collectd-mlab/36)*

stephen-soltesz:
Review status: 0 of 2 files reviewed at latest revision, 9 unresolved discussions.

---

*[site-packages/mlab/disco/simple_session.py, line 1 \[r1\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/36#-KFwyC2Qk5OIFhc4eiSX:-KFyVGwiHk4xWahaexD4:-719267080) ([raw file](https://github.com/m-lab/collectd-mlab/blob/c0626d33ad3e68860c00a697e638b2c254b401d3/site-packages/mlab/disco/simple_session.py#L1)):*
Updated.

---

*[site-packages/mlab/disco/simple_session.py, line 22 \[r1\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/36#-KFx0-CUEwVICwj90NUU:-KFyeo9peXgitoepW1yz:-890468207) ([raw file](https://github.com/m-lab/collectd-mlab/blob/c0626d33ad3e68860c00a697e638b2c254b401d3/site-packages/mlab/disco/simple_session.py#L22)):*
This is a good idea. The error handling / reporting of netsnmp is fairly poor. The following factors prompted me to try to write what I considered a more straight forward and consistent interface.

The two errors we care most about are:
 - connection error, e.g. Timeout
 - unknown or missing OID

netsnmp *does* appear to set ErrorStr for Timeout. So raising an exception for this case makes sense to me. I've added `SNMPError` for this case.

But, netsnmp reports nothing in ErrorStr for unknown or missing OIDs. And, the behavior for `get` and `walk` are different in the "unknown or missing OID" case.

While a VarList is passed to both get and walk, the contents of the list changes after a call to `walk`. On success, elements are replaced or added. On error, the elements are removed entirely (empty list). For `get`, the content remains the same, except on error the Varbind object still has an uninitialized `val` attribute.

My thinking was to make the error case for get and walk look the same from the caller's perspective, i.e. an empty list.

Since netsnmp doesn't set ErrorStr for this case, I'm didn't think raising an exception would be as helpful to the caller.

What do you think? What would be more helpful?

---

*[site-packages/mlab/disco/simple_session.py, line 25 \[r1\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/36#-KFwyWTujXkeUdV1NUYp:-KFz8YyPYNAksd40AWQ1:-1131981712) ([raw file](https://github.com/m-lab/collectd-mlab/blob/c0626d33ad3e68860c00a697e638b2c254b401d3/site-packages/mlab/disco/simple_session.py#L25)):*
Hmm, I imagined that a dict was more generic and simpler than a custom type. If we want to use actual objects, we could use `netsnmp.Varbind` objects, but I also wanted to isolate users from netsnmp.

Can you say more about the value of a custom class over dict?

---

*[site-packages/mlab/disco/simple_session.py, line 57 \[r1\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/36#-KFx384bLQXI1WaSVbHr:-KFz07g02sqk5YZZnEjG:1629328000) ([raw file](https://github.com/m-lab/collectd-mlab/blob/c0626d33ad3e68860c00a697e638b2c254b401d3/site-packages/mlab/disco/simple_session.py#L57)):*
You're right. In my code and communication I've been using "mib" interchangeably with "oid", but they are different things.

My best understanding right now is that "mib" is analogous to "database" and "oid" is analogous to "index to a row of database".

I've updated the instance variable to `_mib` (singular) and parameters to from `mib` to `oid` to be more consistent.

---

*[site-packages/mlab/disco/simple_session.py, line 59 \[r1\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/36#-KFx0VeEZIUR2c1w822C:-KFz2EKXiQrun4pEQUvy:1757116384) ([raw file](https://github.com/m-lab/collectd-mlab/blob/c0626d33ad3e68860c00a697e638b2c254b401d3/site-packages/mlab/disco/simple_session.py#L59)):*
I think you're right. Changed to 'tag'.

---

*[site-packages/mlab/disco/simple_session.py, line 64 \[r1\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/36#-KFx2rjhENF3d23E4bKK:-KFz2Hxqr1D0EVjHMu9L:-223655000) ([raw file](https://github.com/m-lab/collectd-mlab/blob/c0626d33ad3e68860c00a697e638b2c254b401d3/site-packages/mlab/disco/simple_session.py#L64)):*
Also correct. I was using plain dicts at one point, but neglected to update the usage of `.get` after changing to `defaultdict`.

---

*[site-packages/mlab/disco/simple_session_test.py, line 14 \[r1\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/36#-KFx28cH3ZqGoT3CeRZO:-KFzAzYdTjMZ8DYrAr4U:1802517229) ([raw file](https://github.com/m-lab/collectd-mlab/blob/c0626d33ad3e68860c00a697e638b2c254b401d3/site-packages/mlab/disco/simple_session_test.py#L14)):*
Good question. I thought there would be a conflict with creating a real object and simultaneously mocking the `netsnmp.Varbind` object. But, it turns out that there is no conflict, and using a real object is better because I was missing part of the 'tag' value returned by `SimpleSession.get` and `walk`.

---

*[site-packages/mlab/disco/simple_session_test.py, line 39 \[r1\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/36#-KFx1Zbae8E-5nQi8HAy:-KFzE6SyMypGIONZ4KD9:940059655) ([raw file](https://github.com/m-lab/collectd-mlab/blob/c0626d33ad3e68860c00a697e638b2c254b401d3/site-packages/mlab/disco/simple_session_test.py#L39)):*
Yes. There are two error case tests now: empty list return value, and for `SNMPError`.

---

*[site-packages/mlab/disco/simple_session_test.py, line 52 \[r1\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/36#-KFx0zpNc0lSMc0jekgb:-KFzN59JWy52uIhMRwGQ:-1490275066) ([raw file](https://github.com/m-lab/collectd-mlab/blob/c0626d33ad3e68860c00a697e638b2c254b401d3/site-packages/mlab/disco/simple_session_test.py#L52)):*
Yes, I was lazy with the walk test. I've updated `test_walk` to return multiple variables (which is typical) and added the empty-list and `SNMPError` cases.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/collectd-mlab/36)*

mtlynch:
Review status: 0 of 2 files reviewed at latest revision, 5 unresolved discussions.

---

*[site-packages/mlab/disco/simple_session.py, line 22 \[r1\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/36#-KFx0-CUEwVICwj90NUU:-KFzZq2BGA5H1acErOE0:1244737656) ([raw file](https://github.com/m-lab/collectd-mlab/blob/c0626d33ad3e68860c00a697e638b2c254b401d3/site-packages/mlab/disco/simple_session.py#L22)):*
(optional) Empty list feels like too subtle a way to signal an error, especially if a connection error raises an exception now. Could we raise an `SNMPError` for both the `get` and `walk` case for unknown or missing OID that's like `SNMPError('unknown netsnmp error')` or maybe `SNMPError('unknown netsnmp error, OID %s may be invalid' % oid)`

---

*[site-packages/mlab/disco/simple_session.py, line 25 \[r1\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/36#-KFwyWTujXkeUdV1NUYp:-KFzYNf7XcsKqMFCLfVY:-238091443) ([raw file](https://github.com/m-lab/collectd-mlab/blob/c0626d33ad3e68860c00a697e638b2c254b401d3/site-packages/mlab/disco/simple_session.py#L25)):*
Yeah, I should have tagged this one as (optional) as I don't feel *that* strongly, about it, but I do prefer objects. The biggest case I feel dictionaries are confusing are in docstrings because it's much easier to say (and easier for the reader to understand) "a list of netsnmp.Varbind objects" than "a list of dictionaries with the following key, value pairs: ..."

I understand that we're violating encapsulation by using `netsnmp.Varbind` because the promise of this interface is that the user doesn't have to worry about the `netsnmp` APIs. Another way around this might be to write a wrapper object like `simple_session.Varbind`.

---

*[site-packages/mlab/disco/simple_session.py, line 73 \[r2\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/36#-KFz_qBEuWL3nOGLp6oZ:-KFz_qBEuWL3nOGLp6o_:-1273977647) ([raw file](https://github.com/m-lab/collectd-mlab/blob/81dd0a1d2e0cf348b7059530ddb6498cb5bfb19d/site-packages/mlab/disco/simple_session.py#L73)):*
I think we've exceeded complexity for readable list comprehensions in this case:

>Each portion must fit on one line: mapping expression, for clause, filter expression.

https://google.github.io/styleguide/pyguide.html?showone=List_Comprehensions#List_Comprehensions

---

*[site-packages/mlab/disco/simple_session_test.py, line 14 \[r2\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/36#-KFzXcarTjcA_rsgV6ih:-KFzXcarTjcA_rsgV6ii:-238014267) ([raw file](https://github.com/m-lab/collectd-mlab/blob/81dd0a1d2e0cf348b7059530ddb6498cb5bfb19d/site-packages/mlab/disco/simple_session_test.py#L14)):*
Can we do `netsnmp.Varbind(tag, val=val)`? Actually at that point we probably don't even need the helper function.

---

*[site-packages/mlab/disco/simple_session_test.py, line 28 \[r2\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/36#-KFzaOnE_EmdVha7_jZR:-KFzaOnFeTGeePGTLFn-:142191357) ([raw file](https://github.com/m-lab/collectd-mlab/blob/81dd0a1d2e0cf348b7059530ddb6498cb5bfb19d/site-packages/mlab/disco/simple_session_test.py#L28)):*
This is only used in one test right? Suggest defining it there rather than in `setUp` so that the reader doesn't have to jump around to understand that test.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/collectd-mlab/36)*

stephen-soltesz:
Review status: 0 of 2 files reviewed at latest revision, 5 unresolved discussions, some commit checks failed.

---

*[site-packages/mlab/disco/simple_session.py, line 22 \[r1\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/36#-KFx0-CUEwVICwj90NUU:-KFzghHtJpkDLmlZvrH9:1802502753) ([raw file](https://github.com/m-lab/collectd-mlab/blob/c0626d33ad3e68860c00a697e638b2c254b401d3/site-packages/mlab/disco/simple_session.py#L22)):*
Okay. The code that's not been reviewed yet that calls `session.get` or `session.walk` was checking for the empty result and raising an exception anyway. This should be very nearly equivalent. So, Done.

Sometimes there is more context at the call site than within the called function, which is helpful for logging or error reporting. But, in these cases, I think it's okay to raise the exception here.

---

*[site-packages/mlab/disco/simple_session.py, line 25 \[r1\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/36#-KFwyWTujXkeUdV1NUYp:-KFzpM8bPkdTdM_5FPYm:-1712246702) ([raw file](https://github.com/m-lab/collectd-mlab/blob/c0626d33ad3e68860c00a697e638b2c254b401d3/site-packages/mlab/disco/simple_session.py#L25)):*
Okay. I understand. I agree that providing an object interface is easier to document. At this point, my feeling is that the netsnmp library is actually low quality. Ultimately, we would like to use a library like http://easysnmp.readthedocs.org/en/latest/, which has the rational interface that it sounds like we're both imagining.

So, maybe in the spirit of creating a minimal interface that looks like the one we actually want, I've added an `SNMPVariable` object as used by the easysnmp library. And, added TODOs to remove them when easysnmp is available.

How does it look?

---

*[site-packages/mlab/disco/simple_session.py, line 73 \[r2\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/36#-KFz_qBEuWL3nOGLp6oZ:-KFzfvWW0xgePO0KLLFU:-90915138) ([raw file](https://github.com/m-lab/collectd-mlab/blob/81dd0a1d2e0cf348b7059530ddb6498cb5bfb19d/site-packages/mlab/disco/simple_session.py#L73)):*
lol, now that you say so, it clearly has. Done.

---

*[site-packages/mlab/disco/simple_session_test.py, line 14 \[r2\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/36#-KFzXcarTjcA_rsgV6ih:-KFzdCGtwHYia-k2WVwq:-493625192) ([raw file](https://github.com/m-lab/collectd-mlab/blob/81dd0a1d2e0cf348b7059530ddb6498cb5bfb19d/site-packages/mlab/disco/simple_session_test.py#L14)):*
Oh! I missed this possibility. Done.

---

*[site-packages/mlab/disco/simple_session_test.py, line 28 \[r2\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/36#-KFzaOnE_EmdVha7_jZR:-KFzdJHe4jQDEEya5-l9:936219926) ([raw file](https://github.com/m-lab/collectd-mlab/blob/81dd0a1d2e0cf348b7059530ddb6498cb5bfb19d/site-packages/mlab/disco/simple_session_test.py#L28)):*
Yes, I didn't like separating the setup of these values from the use in the assert check. I didn't like the alternatives much either :-)

This is a case when the mock patches interfere with the test arrange block. If we call netsnmp.Varbind within a test then we're calling the mocked version.

Options I see to move the var_list definition closer to its use:

 * move the `@mock.patch.object(netsnmp, 'Varbind')` decorator from the test class to individual functions, then use the context form of patch within `test_walk`. i.e.

        def test_walk(self, mock_varlist):
            var_list = [...]

            with mock.patch.object(netsnmp, 'Varbind'):
                actual = self.session.walk('...')

                self.assertItemsEqual(...)

 * define a new test class for `walk` tests so `setUp` is closer to usage.

* just move the `walk` tests first so that `walk` immediately follows `setUp`.

 * something else I'm not thinking of.

What do you think?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/collectd-mlab/36)*

mtlynch:
Reviewed 2 of 2 files at r3.

---

*[site-packages/mlab/disco/simple_session.py, line 25 \[r1\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/36#-KFwyWTujXkeUdV1NUYp:-KG--ZeaiLzKYgAgeyRI:507749641) ([raw file](https://github.com/m-lab/collectd-mlab/blob/c0626d33ad3e68860c00a697e638b2c254b401d3/site-packages/mlab/disco/simple_session.py#L25)):*
I like it!

---

*[site-packages/mlab/disco/simple_session_test.py, line 28 \[r2\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/36#-KFzaOnE_EmdVha7_jZR:-KG-50TuySsRl0iJKCez:-901296383) ([raw file](https://github.com/m-lab/collectd-mlab/blob/81dd0a1d2e0cf348b7059530ddb6498cb5bfb19d/site-packages/mlab/disco/simple_session_test.py#L28)):*
Oh I completely overlooked that we're mocking `Varbind` and `Varlist`. This seems strange since they're types. They have a little bit of logic, but they're mostly dumb structs. Can we get rid of these mocks and limit our mocking to `netsnmp.Session` APIs?

I think the current mocking actually tests implementation details rather than interfaces. For example, we're asserting this:

```
mock_varbind.assert_called_with('sysDescr.0')
```

because our implementation is like this:

```
oids = netsnmp.VarList(netsnmp.Varbind(oid))
```

But my reading of [Varbind](https://github.com/haad/net-snmp/blob/master/python/netsnmp/client.py#L52) is that the following implementation is also valid, but would break our tests:

```
var = netsnmp.Varbind()
var.tag = oid
oids = netsnmp.VarList(var)
```

If we abstain from mocking these types, we don't run into that issue.

Those types don't define equality operators, so we might need to define custom `self.assertVarbindEqual` methods.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/collectd-mlab/36)*

stephen-soltesz:
Review status: 1 of 2 files reviewed at latest revision, 1 unresolved discussion.

---

*[site-packages/mlab/disco/simple_session_test.py, line 28 \[r2\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/36#-KFzaOnE_EmdVha7_jZR:-KGD7U1LR2-Drm-snPAY:-1281452344) ([raw file](https://github.com/m-lab/collectd-mlab/blob/81dd0a1d2e0cf348b7059530ddb6498cb5bfb19d/site-packages/mlab/disco/simple_session_test.py#L28)):*
Ah ha. Where there's a will, there's a way.

Because `netsnmp.Session.get` and `walk` modify the given `VarList` (and we are mocking `netsnmp.Session` so that it cannot act on the real `VarList` value), I thought we needed to at least control the return value of `netsnmp.VarList`.

It turns out we can redefine the `get` and `walk` functions on demand to modify the varlist as the real versions might.

How does it look now?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/collectd-mlab/36)*

mtlynch:
:lgtm:

---

Review status: 1 of 2 files reviewed at latest revision, 1 unresolved discussion.

---

*[site-packages/mlab/disco/simple_session_test.py, line 28 \[r2\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/36#-KFzaOnE_EmdVha7_jZR:-KGDYL618inJZ88-kB-0:432021016) ([raw file](https://github.com/m-lab/collectd-mlab/blob/81dd0a1d2e0cf348b7059530ddb6498cb5bfb19d/site-packages/mlab/disco/simple_session_test.py#L28)):*
I like it

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/collectd-mlab/36#-:-KGDYNq6VPCdossLmS9d:1417088721)*

mtlynch:
Reviewed 1 of 1 files at r5.
Review status: all files reviewed at latest revision, 1 unresolved discussion, all commit checks successful.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/collectd-mlab/36)*

stephen-soltesz:
Review status: all files reviewed at latest revision, 1 unresolved discussion, all commit checks successful.

---

*[site-packages/mlab/disco/simple_session_test.py, line 28 \[r2\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/36#-KFzaOnE_EmdVha7_jZR:-KGDYfj60HaxfIdnQBYL:1040802091) ([raw file](https://github.com/m-lab/collectd-mlab/blob/81dd0a1d2e0cf348b7059530ddb6498cb5bfb19d/site-packages/mlab/disco/simple_session_test.py#L28)):*
Woohoo!

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/collectd-mlab/36)*

PR #38, opened 2016-04-22
Allow pre-commit example run multiple times

stephen-soltesz:
This small change to allows the pre-commit hook example command to run multiple times. Without this fix, the `rm -rf` will delete the files under the hooks directory. With this fix, the symbolic link is removed instead of the files in the hooks directory.


mtlynch:
LGTM

PR #42, opened 2016-04-26
Also raise exceptions from FakeSession.

stephen-soltesz:
`SimpleSession.get` and `walk` now raise exceptions when no results were returned. So, `FakeSession` should behave the same way.

This PR rearranges some previously private functions to be used by both `SimpleSession` and `FakeSession`.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/collectd-mlab/42)

### DIFF
--- a/site-packages/mlab/disco/simple_session.py
+++ b/site-packages/mlab/disco/simple_session.py
@@ -100,7 +100,7 @@ class FakeSession(object):
     def _find_oid(self, oid):
         result = self._mib[oid]
         if not result:
-            raise SNMPError('netsnmp error unknown; OID may be invalid: ' + oid)
+            raise SNMPError('Empty result set.')
         return result
 
 
@@ -118,7 +118,20 @@ def _varlist_to_list(varlist):
 
 
 def _convert_result(original_oid, varlist):
-    """Converts results in varlist, or raises exceptions for errors."""
+    """Converts a netsnmp.Varlist to a list of SNMPVariable.
+
+    Args:
+        original_oid: str, the original OID requested.
+        varlist: netsnmp.VarList, a container filled with netsnmp.Varbind
+            objects.
+
+    Returns:
+        list of SNMPVariable, each netsnmp.Varbind element from varlist
+            converted to an SNMPVariable.
+
+    Raises:
+        SNMPError: after conversion, the result was an empty set.
+    """
     result = _varlist_to_list(varlist)
     if not result:
         raise SNMPError('netsnmp error unknown; OID may be invalid: ' +
--- a/site-packages/mlab/disco/simple_session.py
+++ b/site-packages/mlab/disco/simple_session.py
@@ -92,10 +92,16 @@ class FakeSession(object):
         self._mib[oid].append(SNMPVariable(tag, value))
 
     def get(self, oid):
-        return _convert_result(oid, self._mib[oid], convert=lambda x: x)
+        return self._find_oid(oid)
 
     def walk(self, oid):
-        return _convert_result(oid, self._mib[oid], convert=lambda x: x)
+        return self._find_oid(oid)
+
+    def _find_oid(self, oid):
+        result = self._mib[oid]
+        if not result:
+            raise SNMPError('netsnmp error unknown; OID may be invalid: ' + oid)
+        return result
 
 
 def _varlist_to_list(varlist):
@@ -111,9 +117,9 @@ def _varlist_to_list(varlist):
     return result
 
 
-def _convert_result(original_oid, varlist, convert=_varlist_to_list):
+def _convert_result(original_oid, varlist):
     """Converts results in varlist, or raises exceptions for errors."""
-    result = convert(varlist)
+    result = _varlist_to_list(varlist)
     if not result:
         raise SNMPError('netsnmp error unknown; OID may be invalid: ' +
                         original_oid)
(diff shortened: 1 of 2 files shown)


mtlynch:
Review status: 0 of 2 files reviewed at latest revision, 1 unresolved discussion.

---

*[site-packages/mlab/disco/simple_session.py, line 98 \[r1\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/42#-KGIjyrVKYArmex694cz:-KGIjyrVKYArmex694d-:113894556) ([raw file](https://github.com/m-lab/collectd-mlab/blob/1c7b98804c54dab0979f35812fd9c2bdf8092615/site-packages/mlab/disco/simple_session.py#L98)):*
I feel like these changes make both classes significantly more difficult to understand. `FakeSession` in particular had both implementations as one-liners (a very nice thing), whereas now the reader has to read 20+ lines to understand the implementation. 

I'm also uncomfortable with sacrificing clarity in the `SimpleSession` to facilitate code reuse in `FakeSession`. I feel like we're sacrificing clarity by requiring `_convert_result` to take a function as a parameter.

Could we keep the implementations independent and change `FakeSession's implementation to something like this?

```
def get(self, oid):
  return self._find_oid(oid)

def _find_oid(self, oid):
  try:
    return self._mib[oid]
  except KeyError:
    raise SNMPError('FakeSession has no value for OID: %s' % oid)
```

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/collectd-mlab/42)*

stephen-soltesz:
Review status: 0 of 2 files reviewed at latest revision, 1 unresolved discussion.

---

*[site-packages/mlab/disco/simple_session.py, line 98 \[r1\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/42#-KGIjyrVKYArmex694cz:-KGIpgPe6AwsOCn7JEN1:1901546991) ([raw file](https://github.com/m-lab/collectd-mlab/blob/1c7b98804c54dab0979f35812fd9c2bdf8092615/site-packages/mlab/disco/simple_session.py#L98)):*
Okay. I don't want to make the classes less clear or harder to understand. I've added the `_find_oid` method to `FakeSession` to keep the implementations separate and removed the "convert" parameter from `_convert_result`.

I think this better preserves the original complexity of both classes, while moving some logic that was previously unnecessarily part of `SimpleSession` to the module level.

How does it look?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/collectd-mlab/42)*

mtlynch:
:lgtm:

---

Reviewed 1 of 2 files at r1, 1 of 1 files at r2.

---

*[site-packages/mlab/disco/simple_session.py, line 103 \[r2\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/42#-KGIsjBm2C34xqmdX1rX:-KGIsjBm2C34xqmdX1rY:-25269053) ([raw file](https://github.com/m-lab/collectd-mlab/blob/8ab5a193173f56231b99a3fa2a4ed1d61e8d2135/site-packages/mlab/disco/simple_session.py#L103)):*
Is there a requirement that the error message be exactly the same in both cases? It seems strange that we'd say it's an unknown error when really we control everything about `FakeSession` and we know why there's no result: the caller never `prepare`d one.

---

*[site-packages/mlab/disco/simple_session.py, line 121 \[r2\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/42#-KGItC-c8Svx06GdWh-x:-KGItC-c8Svx06GdWh-y:-1833401537) ([raw file](https://github.com/m-lab/collectd-mlab/blob/8ab5a193173f56231b99a3fa2a4ed1d61e8d2135/site-packages/mlab/disco/simple_session.py#L121)):*
Can we expand the docstring a bit? Now that it's a little farther away from the related functions, it's harder to keep track of what we're converting to what. I think it'd be helpful to have the args explicitly documented.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/collectd-mlab/42#-:-KGIuOcoK5qD6COYOqYR:1417088721)*

stephen-soltesz:
Review status: all files reviewed at latest revision, 2 unresolved discussions, all commit checks successful.

---

*[site-packages/mlab/disco/simple_session.py, line 103 \[r2\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/42#-KGIsjBm2C34xqmdX1rX:-KGIvKZuMfXWg2AKMVpY:-1021784607) ([raw file](https://github.com/m-lab/collectd-mlab/blob/8ab5a193173f56231b99a3fa2a4ed1d61e8d2135/site-packages/mlab/disco/simple_session.py#L103)):*
There's no requirement that they be the same. You're right that the caller should know what to expect from a `FakeSession`.

Changed the error message to simply "Empty result set."

---

*[site-packages/mlab/disco/simple_session.py, line 121 \[r2\]](https://reviewable.io:443/reviews/m-lab/collectd-mlab/42#-KGItC-c8Svx06GdWh-x:-KGIy-gcvBiZkHRNHl7N:1328085673) ([raw file](https://github.com/m-lab/collectd-mlab/blob/8ab5a193173f56231b99a3fa2a4ed1d61e8d2135/site-packages/mlab/disco/simple_session.py#L121)):*
No problem.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/collectd-mlab/42)*

PR #47, opened 2016-05-05
Disco package integration

stephen-soltesz:
This change includes several new configuration files to integrate the switch traffic collection as a standard part of the collectd-mlab package.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/collectd-mlab/47)


collectd-mlab.spec
```diff
@@ -77,17 +83,41 @@ install -D -m 644 viewer/collection-mlab.conf		%{buildroot}/etc/collection-mlab.
...
+install -d %{buildroot}/%{site_packages}/mlab/disco
+install -D -m 644 site-packages/mlab/disco/__init__.py \
+                  site-packages/mlab/disco/arp.py \
+				  site-packages/mlab/disco/collectd.py \
+                  site-packages/mlab/disco/discovery.py \
+                  site-packages/mlab/disco/models.py \
+                  site-packages/mlab/disco/network.py \
+                  site-packages/mlab/disco/route.py \
+                  site-packages/mlab/disco/simple_session.py \
+				  %{buildroot}/%{site_packages}/mlab/disco
```
nkinkade: Is this extra indent intentional, and the one on line 95?

collectd-mlab.spec
```diff
@@ -77,17 +83,41 @@ install -D -m 644 viewer/collection-mlab.conf		%{buildroot}/etc/collection-mlab.
...
+install -D -m 644 plugin/mlab.py			%{buildroot}/var/lib/collectd/python/mlab_vs.py
+
+# Init script.
+install -D -m 755 plugin/collectd-setup		%{buildroot}/etc/init.d/collectd-setup
+
+# Python site-packages modules.
+install -d %{buildroot}/%{site_packages}/mlab/disco
+install -D -m 644 site-packages/mlab/disco/__init__.py \
+                  site-packages/mlab/disco/arp.py \
+				  site-packages/mlab/disco/collectd.py \
```
nkinkade: Is this extra indent intentional?

collectd-mlab.spec
```diff
@@ -77,17 +83,41 @@ install -D -m 644 viewer/collection-mlab.conf		%{buildroot}/etc/collection-mlab.
...
+                  site-packages/mlab/disco/network.py \
+                  site-packages/mlab/disco/route.py \
+                  site-packages/mlab/disco/simple_session.py \
+				  %{buildroot}/%{site_packages}/mlab/disco
+
+# Disco commands.
+install -D -m 755 disco/disco_config.py	%{buildroot}/usr/bin/disco_config.py
+
+# Disco config.
+install -D -m 755 disco/models.yaml \
```
nkinkade: Does a YAML file need the to be executable?

collectd-mlab.spec
```diff
@@ -77,17 +83,41 @@ install -D -m 644 viewer/collection-mlab.conf		%{buildroot}/etc/collection-mlab.
...
+    %{buildroot}/usr/share/collectd-mlab/models.yaml
```
nkinkade: Can /models.yaml be removed from the end of this destination, since it's the same name as the origin file?

collectd-mlab.spec
```diff
@@ -77,17 +83,41 @@ install -D -m 644 viewer/collection-mlab.conf		%{buildroot}/etc/collection-mlab.
...
 
 #
 # Files for "collectd-mlab-vsys" package.
 #
 
 # Monitoring.
-install -D -m 755 monitoring/check_collectd_mlab.py	%{buildroot}/usr/lib/nagios/plugins/check_collectd_mlab.py
+install -D -m 755 monitoring/check_collectd_mlab.py	\
```
nkinkade: For consistency, can the line continuation character be a single space from the end of the prior text?

collectd-mlab.spec
```diff
@@ -130,6 +179,14 @@ fi
...
 chkconfig rsyslog on
 chkconfig collectd on
+chkconfig collectd-setup on
+
+# TODO(soltesz): fix mlab_export.py to handle initial conditions correctly.
+# Initialize the lastexport timestamp file to one hour ago.
+touch -t $( date +%Y%m%d%H00 -d "-1 hour" ) /var/lib/collectd/lastexport.tstamp
+
+# TODO(soltesz): what package should own this file?
+touch %{site_packages}/mlab/__init__.py
```
nkinkade: Should the path be /mlab/**disco**/\_\_init__.py?

plugin/collectd-setup
```diff
@@ -0,0 +1,70 @@
...
+# chkconfig: - 85 16
+# description: collectd-setup performs auto-discovery of the local switch
+#   configuration to complete collectd setup. collectd-setup should run before
+#   collectd.
+
+
+# NOTE: Perform SNMP auto-discovery for switch configuration.
+# If we cannot update the snmp community string, then the script should halt.
+COLLECTD_SNMP="/etc/collectd-snmp.conf"
+COMMUNITY="/home/mlab_utility/config/snmp.community"
```
nkinkade: Is config/ a new directory? On a running slice right now I see there is conf/, but no config/.

plugin/collectd-setup
```diff
@@ -0,0 +1,70 @@
...
+
+    # /etc/collectd-mlab.conf includes the $COLLECTD_SNMP configuration file
+    # unconditionally. So, we must touch the file to create an empty
+    # configuration even when we cannot generate the SNMP configuration.
+    touch "${COLLECTD_SNMP}"
+}
+
+case "$1" in
+  start)
+        start
```
nkinkade: Is there some particular reason for the non-standard indention in this case statement?  The rest of the file seems to use 4 spaces, but this is a mix of 2 spaces and 8 spaces.

plugin/collectd-setup
```diff
@@ -0,0 +1,70 @@
...
+# NOTE: Perform SNMP auto-discovery for switch configuration.
+# If we cannot update the snmp community string, then the script should halt.
+COLLECTD_SNMP="/etc/collectd-snmp.conf"
+COMMUNITY="/home/mlab_utility/config/snmp.community"
+SWITCH="s1.${HOSTNAME#*.}"
+
+function start() {
+
+    # If the community string file is missing or zero size.
+    if ! test -s "${COMMUNITY}" ; then
```
nkinkade: To be sure, at this stage, disco_config.py will already have been run once to have generated $COMMUNITY? On the surface, it looks as if for some reason $COMMUNITY doesn't exist or is empty, that it will always stay that way without manual intervention. Most likely I'm just not understanding how the RPM install process works fully, or misreading this script.

collectd-mlab.spec
stephen-soltesz: No. All the mixed indents are due to my editor settings not being applied to `*.spec` files. I need to find a way to fix this.

collectd-mlab.spec
stephen-soltesz: No! Good catch. This was a copy/derp error.

collectd-mlab.spec
stephen-soltesz: We use this convention throughout the file. Is this approach unusual in general?

collectd-mlab.spec
stephen-soltesz: Yes. This was another tab character.

collectd-mlab.spec
stephen-soltesz: The file `mlab/disco/__init__.py` is declared in the `%files` section above.

This file `mlab/__init__.py` is needed so that python scripts can import the `mlab.disco` module, For example:

    from mlab.disco import network

But, I put this `touch` here because I didn't think that collectd-mlab should own the `mlab` root directory. The TODO is to find a better owner. I'm thinking the `package` repo should own this, but then the collectd-mlab package has a dependency on it in a weird way. I don't have a better solution yet.

plugin/collectd-setup
stephen-soltesz: I'll use `conf`. I need to update the slice data backup script to use the `conf` path instead.

plugin/collectd-setup
stephen-soltesz: Good question. The logic is a little particular.

The expected order of operations are:

* slice reinstalled
* community string distributed by ansible or restored by slice data backup.
* disco_config generates an updated community string.
* disco_config generates an snmp configuration for collectd.
* collectd collects switch traffic data.

At this point, if the community string has not been distributed yet, then the check `! test -s "${COMMUNITY}"` will be true and the error will be printed. Once it is present, we can proceed to the community string update.

What would help make this clearer?

plugin/collectd-setup
stephen-soltesz: The only reason is my love of mixing tabs and spaces accidentally. Fixed.

plugin/collectd-setup
nkinkade: No, it was just my not recognizing/forgetting that the file $COMMUNITY was originally put in place by Ansible.  I had the mistaken impression when reading this script that $COMMUNITY was created by the script.  Now I understand. Thanks.

nkinkade:
LGTM

PR #51, opened 2016-05-09
Monitor disco configuration and local collection

stephen-soltesz:
This change adds monitoring checks for the switch configuration and a local end to end check that switch data is being collected. The switch configuration check looks for the `snmp.community` file and it's `.updated` counterpart. Finally, we confirm that local discards are collected as an indication that all switch metrics are collected.

If any check fails, the plugin fails with a CRITICAL status.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/collectd-mlab/51)


nkinkade:
LGTM

PR #52, opened 2016-05-10
Add option to export counts rather than rates.

stephen-soltesz:
This PR adds a new boolean flag to `mlab_export.py [--counts]` that exports metric counts rather than rates. The default is False, which preserves current behavior.

This feature is needed to export raw counts for the switch traffic data.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/collectd-mlab/52)


nkinkade:
LGTM

PR #56, opened 2016-06-08
Return VLAN id 1 for Cisco auto discovery.

stephen-soltesz:
Previously we expected that a Cisco switch would be configured either
with no VLANs (defaulting to vlan id 1) or with at least one vlan (not
vlan id 1). In practice we discovered that one site included a configuration
where M-Lab nodes were running on vlan 1 but a second vlan was also
configured unrelated to the M-Lab nodes. This caused the auto discovery
to fail.

To This change returns vlan id 1 and allows such configurations to
succeed.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/collectd-mlab/56)


nkinkade:
LGTM

PR #64, opened 2019-08-14
Make collectd-mlab package compatible with k8s docker container

stephen-soltesz:
These changes break backward compatibility for the collectd-mlab package on the vserver system.

In particular, the collectd-mlab RPM now includes the check_collectd_mlab.py script and a cron job to run it periodically to generate node-exporter prometheus metrics. (copied from https://github.com/m-lab/nodebase-private/blob/master/support/collectd_mlab.cron)

The `mlab` python plugin that monitored vserver metrics is now disabled. Several resource checks in check_collectd_mlab are removed because they were created by the mlab plugin.

Rather than using the system hostname (which can be different from host on k8s) we read the `HOSTNAME` from `/home/mlab_utility/conf/hostname` or the environment.

Finally, the DISCO config generation logic checks for all `eth*` interfaces with IPv4 addresses. Previously, a single interface was assumed, but this is no longer true in k8s. The device must start with "eth". This should work with k8s configs where `hostNetwork: true`.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/collectd-mlab/64)


stephen-soltesz:
https://github.com/m-lab/dev-tracker/issues/269

PR #69, opened 2019-11-06
Adds a new switch model to models.yaml for the Juniper EX4200 

nkinkade:
Currently the only site with an EX4200 is HND01. It will likely be the only site that  ever has an EX4200, and the site is also running Dell R620s. Additionally it's a singleton site. The future of HDN01 is not certain, but for now it is running in production so we need to support it.

One alternative to this solution would be to find a common string in `sysDescr.0` between the QFX and EX, since they share the same config. For example, the word `Juniper` is common between them and we could rename the QFX section as `Juniper`, which could catch both the QFX and EX switches.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/collectd-mlab/69)


nkinkade:
Look at the data with @stephen-soltesz, collectd "worked" when the nodes were PLC nodes, in the sense that collectd did not throw and exception. But all of the data for switch utilization is nothing but zeros. @stephen-soltesz believes that this is likely the result of upgrading the old Procurve switch to a Juniper EX4200 without having reinstalled the nodes:

> I think that could do it.
> 
> b/c there was an old, valid snmp config locally, collectd started successfully, the health checks passed b/c it was looking for any value from a switch metric rather than a non-zero one, and the old rrd files would have provided that. Once the switch was replaced, the probes failed.

